PR #6, opened 2016-09-15
basic endpoint setup

zoeabryant:
- [x] an endpoint that invokes the function
- [x] some form of validation, like a required secret key
- [ ] prismic webhooks configured

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,2 @@
 BUCKET_NAME_PREFIX="my-website-honestly"
 BUCKET_NAME_SUFFIX="dev"
-API_KEY='my-website-honestly'
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,3 @@
 BUCKET_NAME_PREFIX="my-website-honestly"
 BUCKET_NAME_SUFFIX="dev"
+SECRET_AUTH="bunch of generated random characters here"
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -1,2 +1,3 @@
 BUCKET_NAME_PREFIX="my-website-honestly"
 BUCKET_NAME_SUFFIX="dev"
+API_KEY='my-website-honestly'
```
lpil: This should be very clear that it needs to be changed!

lpil: Ah, misunderstanding. It seems this is the name of the key, AWS handles generation for us.

No need for this to be an env var then,

lpil: Also needs to be named after the client that uses it :)

zoeabryant: have clarified on slack that this is for the API_KEY name and definitely not the key itself. Going to move it inline as it's not a secret.

publish-service/index.js
```diff
@@ -15,7 +16,11 @@ function doPublish(cb) {
 
 export function publish(event, context, cb) {
   try {
-    doPublish(cb);
+    if (event && event.body && event.body.secret === secretAuth) {
```
lpil: Do we need to check for event?

zoeabryant: I was conscious of 'cannot get body of undefined' errors - but I think it's highly unlikely to get undefined. You can get an empty body but probably not an actual undefined.
Will remove it.

lpil:
What's going on here Zoe?

asavin:
Closing this one as it is stale.

PR #7, opened 2016-09-15
Dev server

lpil:
- [x] Browser based app for dev work
- [x] Hot module reloading
- [x] Dummy react app
- [x] Router

I've gotta say, hot module reloading is freaking sweet.


Makefile
```diff
@@ -5,13 +5,17 @@ NBIN=./node_modules/.bin
...
+NPM_CHECK_UPDATES=$(NBIN)/ncu
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 
 badger: ## Red Badger 4 lyfe!
+	@echo ''
 	@cat assets/badger.txt
+	@echo ''
```
zoeabryant: can you add these blank lines to the file itself?

lpil: I thought maybe it would be annoying if you have to trim the lines if you wanted it without it? I dunno.

lpil:
Nope, need to do that next.

Makefile
zoeabryant: We can cross that bridge if we need to :) I just think it'd look tidier in the file itself.
Of course, this doesn't block the PR going through

PR #8, opened 2016-09-15
Local dev compiler

lpil:
This implements the static site compiler.

The local dev compiler has been tested.

The lambda version may work as a side effect, but this has not been tested and is outside the scope of this PR.


AndrewBestbier:
I think 'Speel moore gud' is one of the more unique commit messages I have seen :)

AndrewBestbier:
'Avoid continuing after rejection'. Profound. Okay I'll stop

lib/array/index.js
```diff
@@ -0,0 +1,11 @@
+export function flatten(array) {
+  return Array.prototype.concat.apply([], array);
+}
+
+export function flatMap(array, lambda) {
+  return flatten(array.map(lambda));
+}
+
+export function uniq(array) {
```
asavin: Why not just pick these from Ramda?

site/compiler/index.js
```diff
@@ -1,14 +1,36 @@
...
-  },
-];
-
-module.exports = function compileSite() {
-  return pages;
-};
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import { match, RouterContext } from 'react-router';
+import { uniq } from '../../lib/array';
```
asavin: Again not sure what's the benefit of our own library for handling arrays. We'll likely eventually have switch to something like Ramda anyway, and then will have to remember to refactor existing code.

site/compiler/parse-router/index.js
```diff
@@ -0,0 +1,45 @@
+import { flatMap } from '../../../lib/array';
+
+function fullPath(route, prefix) {
+  return (prefix + (route.props.path || '')).replace(/\/+/, '/');
```
zoeabryant: I find it quite difficult to read this function and understand what is going on. Would you mind adding a comment to explain?

site/compiler/parse-router/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+function routeInfo(route, prefix) {
+  return {
+    path: fullPath(route, prefix),
+    numChildren: (route.props.children || []).length,
+    indexRoute: route.type.displayName === 'IndexRoute',
+  };
+}
+
+function getChildren(route) {
```
zoeabryant: what does children refer to in this context?

lib/array/index.js
lpil: They were so small it was quicker to write them rather than break flow and look at the docs for ramda. I can switch them out now I've finished.

site/compiler/parse-router/index.js
lpil: I don't like comments as they go stale. I'll refactor the code to be clearer.

site/compiler/parse-router/index.js
lpil: The children of the route. It's React :)

zoeabryant: Discussed on slack, this is for nested 'Route' components in JSX

AndrewBestbier:
Looks good Louis :)

PR #12, opened 2016-09-21
CSS modules

asavin:
giphy 8


package.json
```diff
@@ -45,5 +51,7 @@
         "make test"
       ]
     }
+  },
+  "devDependencies": {
```
asavin: I'm sort of puzzled why all webpack loaders are now required by eslint to be placed in the common dependencies instead of devDeps?

lpil:
I think the linter is broken in some fashion. It's requiring different formatting though we're using the same rules.

webpack.config.js
```diff
@@ -11,6 +12,10 @@ const baseConfig = {
   module: {
     loaders: [
       {
+        test: /\.css$/,
+        loader: ExtractTextPlugin.extract('style-loader', 'css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!postcss-loader'),
```
lpil: Can we discard the CSS when building the lambda?

asavin: Sure

asavin:
This is currently broken when trying to serve with `make dev` - the error i'm getting is `sha1.js:36Uncaught TypeError: inherits is not a function` from `sha1.js`. Something something webpack broken

webpack.config.js
asavin: Oh i see - move CSS compilation to `devAppConfig`? That would be for dev - how do we enable CSS compilation when on lambda?

asavin: What's the assets releasing process? I'm vaguely remembering we spoke about compiling and pushing all assets (css, images) as part of a lambda deploy to a separate S3 bucket.

asavin:
And fixed.

package.json
```diff
@@ -9,21 +9,28 @@
...
     "json-loader": "^0.5.4",
     "mocha": "^3.0.2",
+    "node-hook": "^0.4.0",
     "npm-check-updates": "^2.8.0",
+    "postcss-loader": "^0.13.0",
     "pre-git": "^3.10.0",
     "ramda": "^0.22.1",
     "react": "^15.3.1",
     "react-dom": "^15.3.1",
+    "react-helmet": "^3.1.0",
```
zoeabryant: 👍 this looks like a nice module!

webpack.config.js
zoeabryant: I believe that's the ideas, and that the deployment would be triggered by CircleCi

asavin: Ok cool - let's leave it as is for now - this should allow us to carry on with the user story while developing locally. We can figure out the rest of the deploy process once everyone are back.

package.json
lpil: Cool.

PR #24, opened 2016-09-30
Added divider style

zoeabryant:
### Motivation
gif
Creating the angle that is used frequently in [the designs](zpl.io/Z2lYbwG)

This is CSS only version.

I looked at how to do this with mixins [with this package](https://github.com/postcss/postcss-mixins) but it seemed to require a whole bunch of other stuff (simple vars, nesting) to use. It seemed unnecessary.

### Test plan
- Run it, see the angles.
- It is responsive
- do cross browser checks

### Pre-flight checklist
- [ ] Has been design reviewed (!)
- [x] Add a gif


site/css/_divider.css
```diff
@@ -0,0 +1,30 @@
...
+
+.divider:before {
+  content:"";
+  position: absolute;
+  right: -6px;
+  width: 50%;
+  height: 10px;
+  top: 0px;
+  -webkit-transform: skew(-40deg);
+  -moz-transform: skew(-40deg);
```
asavin: don't we have autoprefixer for this?

PR #31, opened 2016-10-04
Adding primary button shared class, corrected example in file

zoeabryant:
A button component for us all to use in our various tasks:

('scuse the context: it's an example usage)


zoeabryant:
note: this has not been design reviewed, and will need to be expanded in the future to have a hover state, maybe a text-shadow.
this is just so we all work with the same class

PR #32, opened 2016-10-04
Serving Sentinel font from typography dot com servers

asavin:
tumblr_inline_odd3fcgwaw1raprkq_500

### Motivation

Unlike other font providers, Typography.com requires you to serve fonts in a _very specific_ way. Namely they want to decide which font your client needs on their backend. On development they would then serve that font from their backend (yes, this means you'll always need internet connection), and on prod they would point the client to the correct asset hosted on our server.

### Test plan
- Check that serif fonts are displayed correctly on all browsers.

### For production

We will have to create `[production-url]/fonts` folder, upload a bunch of css files in there and let typography.com know about that - that should enable production serving of the fonts. The magic would still happen on their backend, but the actual assets would be served from our servers. There is obv an issue that we will not be able to minify everything into a single css file, otherwise typography.com servers will explode.

### Pre-flight checklist
- [x] Has been design reviewed
- [x] Add a gif


asavin:
Font serving now also works for preview apps

Issue #50, opened 2016-10-10
homepage - iphone 7 - iOS 10 - page scrolls horizontally


Roisi:
On the site on iphone 7, the page has a slight horizontal scroll as well as vertical

Roisi:
this was on Safari

zoeabryant:
related to #35

zoeabryant:
duplicate - this is caused by #35

Issue #35, opened 2016-10-04
Divider: Kink appearing next to the divider on the side on iPhone 5S


sarigriffiths:
When you scroll horizontally.

![kink appearing on the right of divider](https://cloud.githubusercontent.com/assets/3176873/19081511/73669486-8a52-11e6-87d9-00d4f98b8242.jpg)

zoeabryant:
related to #50

PR #58, opened 2016-10-11
Safari requires position: relative; to be set when applying overflow-x hidden

zoeabryant:
### Motivation
#50 , #35
It was also happening in safari on OSX

[fix credit](http://www.tonylea.com/2010/safari-overflow-hidden-problem/)

### Test plan
- View in safari on iOS devices.
- on iPhone 5S and 7 in particular

### Pre-flight checklist
- n/a Unit tests
- n/a Has been design reviewed
- `¯\_(ツ)_/¯` Add a gif

### DIFF
--- a/site/layout/style.css
+++ b/site/layout/style.css
@@ -7,5 +7,4 @@
 html, body {
   overflow-x: hidden;
   position: relative;
-  background: white;
 }
--- a/site/layout/style.css
+++ b/site/layout/style.css
@@ -7,4 +7,5 @@
 html, body {
   overflow-x: hidden;
   position: relative;
+  background: white;
 }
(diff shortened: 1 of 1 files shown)


site/layout/style.css
```diff
@@ -4,6 +4,8 @@
   box-sizing: border-box;
 }
 
-body {
+html, body {
   overflow-x: hidden;
+  position: relative;
+  background: white;
```
lpil: What is the `background: white` for?

zoeabryant: discussed in person

zoeabryant:
in @AndreiDanielIonescu 's hands now :)

Issue #104, opened 2016-10-24
Handle patchy networks and display our own error instead of the Dinosaur


asavin:
![giphy 7](https://cloud.githubusercontent.com/assets/487468/19642445/3812f5ec-99dd-11e6-8c41-97a3c9637511.gif)

When network is patchy and RB site has already loaded at least once, we should be able to take over the network errors and display our own error page with navigation instead of letting it fail into the browser's own empty page.

lpil:
-1
I love the dinosaur

tabazevedo:
Closing as this was handled in https://github.com/redbadger/website-honestly/pull/223

PR #108, opened 2016-10-25
Sw

lithin:
### Motivation

- #104 step int he right direction - if a user has loaded any page at least twice, they'll never see the dinosaur again. That is, if they us a hyper-modern browser :D

### Test plan

- in Chrome and Firefox: load any website honestly page at least twice; then go offline and test that it still loads
- check that all assets still load fine in all other browsers

### Pre-merge checklist

- [ ] Unit tests
- [x] Reviews
  - [x] Code review
  - [x] Design review N/A
- [ ] Manual testing
  - [x] Windows 7 IE 11 N/A
  - [x] Windows 10 Edge N/A
  - [x] Mac OS El Capitan Safari N/A
  - [x] Mac OS El Capitan Firefox (supports SW but doesn't work offline)
  - [x] iOS 9 Safari N/A
  - [ ] Android 6 Chrome (should work fully but can't test it via browserstack)
- [ ] Tester approved

### DIFF
--- a/bin/ci-deploy.sh
+++ b/bin/ci-deploy.sh
@@ -35,7 +35,7 @@ deployMaster() {
   echo Copying assets to S3
   aws s3 sync ./dist/assets-honestly s3://$BUCKET_NAME/assets-honestly
   aws s3 sync ./dist/txt s3://$BUCKET_NAME
-  aws s3 cp ./dist/sw.js s3://$BUCKET_NAME
+  aws s3 sync ./dist/sw.js s3://$BUCKET_NAME
   make services-deploy
   make publish-service-invoke
   echo Done!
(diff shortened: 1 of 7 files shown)


bin/ci-deploy.sh
```diff
@@ -34,6 +35,7 @@ deployMaster() {
   echo Copying assets to S3
   aws s3 sync ./dist/assets-honestly s3://$BUCKET_NAME/assets-honestly
   aws s3 sync ./dist/txt s3://$BUCKET_NAME
+  aws s3 cp ./dist/sw.js s3://$BUCKET_NAME/$COMMIT_REF
```
zoeabryant: no ref

site/compiler/asset-paths.js
```diff
@@ -10,3 +10,5 @@ const assets = assetsDigest[bundleName];
 
 export const cssPath = filePath(assets.css);
 export const jsPath = filePath(assets.js);
+
+export const HI = 'HI';
```
zoeabryant: gone

lpil:
Does this work for other pages or just the index?

lithin:
Works everywhere! :)

asavin:
Since it is impossible to test it on preview app because of the SHA in the url, should we merge this, test on staging and see if it works?

lpil:
It's not impossible, just import the path prefix the same way that the router does :)

PR #122, opened 2016-10-28
Add Flow type checker

lpil:
### Motivation

In an attempt to avoid the same problems on RL with unreliable data and
missing edge cases causing crashes I'm adding Flow. :)
### Pre-merge checklist
- [x] Reviews
  - [x] Code review


asavin:
I'd say this will become relevant once we have Prismic related tickets on the backlog. Probably a good idea, but not yet relevant - the dynamic part of this site might never happen.

lpil:
It'll be hard to reteroactively add this. I think it's better to do sooner than later.

lithin:
Yeah the sooner this is in, the better - takes quite some time to get used to it

lpil:
@AndrewBestbier @zoeabryant Any opinions? :)

zoeabryant:
As @asavin will be maintaining the project going forward, I think it's his call to make.

asavin:
Going to de-scope this for now, let's get back to this after launch next week.

JoeStanton:
Reopened this after discussion yesterday. I've added a couple of basic annotations, and it should largely work out of the box with anyone using Nuclide/Atom.

lpil:
<£

Makefile
```diff
@@ -46,6 +46,10 @@ test: ## Run the tests
 	$(MOCHA)
 
 
+flow: ## Run the type checker
```
lpil: Add to PHONY

PR #133, opened 2016-11-01
Fix SW

lithin:
### Motivation

Service Workers didn't work on "What We Do" - now they do!

### Test plan

- [ ] Service workers work on What We Do page
- [ ] What We Do page gets cached as well as the home page

### Pre-merge checklist

- [x] Unit tests N/A
- [x] Reviews
  - [x] Code review
  - [x] Design review N/A
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved

### DIFF
--- a/site/client/index.js
+++ b/site/client/index.js
@@ -15,7 +15,9 @@ export function makeApp({ element, data, history }) {
   registerStateNavigator(stateNavigator);
 
   if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.register('/sw.js');
+    console.log('hey there');
+    console.log(location.origin);
+    navigator.serviceWorker.register(location.origin + '/sw.js');
   }
 
   routes.forEach(route => {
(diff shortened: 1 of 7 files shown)


site/sw.js
```diff
@@ -7,6 +7,7 @@ self.addEventListener('install', event => {
     caches.open(CACHE_NAME).then(cache => {
       return cache.addAll([
         '/',
+        'what-we-do/',
```
lpil: Seems like we could build this list from the router.

site/client/index.js
```diff
@@ -15,7 +15,9 @@ export function makeApp({ element, data, history }) {
   registerStateNavigator(stateNavigator);
 
   if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.register('sw.js');
+    console.log('hey there');
+    console.log(location.origin);
```
lpil: Logging left in here.

site/sw.js
lpil: Why is that? With webpack it should compile to a normal function call.

PR #150, opened 2016-11-03
Display the footer map SVG with text in place

zoeabryant:
Text is missing on live!
& removing some unused images


zoeabryant:
now in #151

PR #149, opened 2016-11-03
Display the map pin in the footer

AndrewBestbier:
### Motivation

- [Related story](http://...)

TODO: write a motivation.

### Test plan

- TODO: Write a test plan.

### Pre-merge checklist

- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] Mac OS El Capitan Safari
  - [ ] Mac OS El Capitan Chrome
  - [ ] Mac OS El Capitan Firefox
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved

PR #151, opened 2016-11-03
Footer map fixes

zoeabryant:
**Fixes in this PR:**

- [x] map pin missing in footer (no longer #149)

- [x] Replace missing map text (no longer #150)

- [x] Google maps link should open in new tab

... and a slight deviation of deleting unused assets


site/components/footer/index.js
```diff
@@ -143,6 +143,8 @@ const Footer = () => (
 
           <a
             className={styles.mapLink}
+            target="_blank"
+            rel="noopener noreferrer"
```
zoeabryant: eslint complained without this, see blog post: https://mathiasbynens.github.io/rel-noopener/

PR #175, opened 2016-11-04
Require valid email for contact us form

lithin:
### Motivation

In order to make our Contact Us form more resilient, we decided to require an email when users are submitting a Contact Us form. In other words, @AndreiDanielIonescu was right :D

### Test plan

- Check you can send email via contact us form
- Check you can't do it if there isn't a valid email submitted

### Pre-merge checklist

- [x] Documentation (does it need any?)
- [x] Unit tests
- [ ] Reviews
  - [x] Code review
  - [x] Design review N/A
- [x] Manual testing (probably not necessary as it is mainly back-end/js change?)
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved

### DIFF
--- a/services/contact_us.js
+++ b/services/contact_us.js
@@ -2,13 +2,14 @@
 import aws from 'aws-sdk'; // eslint-disable-line import/no-unresolved, import/no-extraneous-dependencies
 import { validateAndSendEmail } from './ses';
 
-const contactUsEmailTo = process.env.CONTACT_US_EMAIL_TO;
-
-if (!contactUsEmailTo) {
-  throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
-}
 
 export default function doContactUs(event, cb) {
+  const contactUsEmailTo = process.env.CONTACT_US_EMAIL_TO;
+
+  if (!contactUsEmailTo) {
+    throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
+  }
+
   const data = {
     emailTo: contactUsEmailTo,
     message: event.body.message,
--- a/services/ses/index.js
+++ b/services/ses/index.js
@@ -4,7 +4,7 @@ const contactLabel = 'Contact details:';
 const fieldMessages = {
   emailTo: 'Must be present',
   message: 'How can we help? Let us know in the box below.',
-  contact: 'Please let us know the best way of contacting you.',
+  contact: 'Please let us know your email address.',
 };
 
 function filterInternalErrors(error) {
@@ -15,6 +15,11 @@ function filterInternalErrors(error) {
     };
   }
   throw error;
+}
+
+function validateContact(contact) {
+  const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/; // eslint-disable-line max-len
+  return re.test(contact);
 }
 
 function throwValidationError(errors) {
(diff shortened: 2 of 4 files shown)


services/contact_us.js
```diff
@@ -2,13 +2,14 @@
...
-  throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
-}
 
 export default function doContactUs(event, cb) {
+  const contactUsEmailTo = process.env.CONTACT_US_EMAIL_TO;
+
+  if (!contactUsEmailTo) {
+    throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
+  }
+
```
zoeabryant: 👍

services/ses/index.js
```diff
@@ -4,7 +4,7 @@ const contactLabel = 'Contact details:';
 const fieldMessages = {
   emailTo: 'Must be present',
   message: 'How can we help? Let us know in the box below.',
-  contact: 'Please let us know the best way of contacting you.',
+  contact: 'Please let us know your email address.',
```
zoeabryant: are we using this message for validation errors as well?

services/ses/index.js
```diff
@@ -17,6 +17,11 @@ function filterInternalErrors(error) {
   throw error;
 }
 
+function validateContact(contact) {
+  const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/; // eslint-disable-line max-len
```
zoeabryant: where did this come from, out of curiosity?

services/contact_us.js
```diff
@@ -2,13 +2,14 @@
...
-if (!contactUsEmailTo) {
-  throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
-}
 
 export default function doContactUs(event, cb) {
+  const contactUsEmailTo = process.env.CONTACT_US_EMAIL_TO;
+
+  if (!contactUsEmailTo) {
+    throw new Error('CONTACT_US_EMAIL_TO environment variable not set!');
+  }
```
lpil: This should be outside the function as it only needs to run once when node loads the code, not every time the function is called.

services/ses/index.js
lpil: This is pretty grim and probably incorrect. I'd use something simpler like `/.+@.\..+/`

services/contact_us.js
zoeabryant: ^ we were having a (possibly separate issue) that when that Env variable wasn't set, it was showing up on the front end when a user tries to use the mailing list form. Moving this into this function would prevent that issue.

lpil: That's an unrelated issue. That needs to be fixed on the client- i.e. don't pass unverified content from the API to the client.

services/ses/index.js
lpil: Oh wow, this explicitly allows IPs (`whatever@192.235.23.45`). That's pretty weird.

jameshopkins: I don't think you can _reliably_ validate an email address

zoeabryant: What do we use for the mailing list email input? I think they should be the same, it would be a bit strange if an email passed in one case but not the other.

lpil: The other is `/.+@.+\..+/`

services/contact_us.js
zoeabryant: raised in #176 to make sure it gets addressed

services/ses/index.js
MarcelCutts: (Beware, `regex + emails` lead to madness. To be fair, you can replace `emails` with pretty much anything and have this still be a true statement.)

lpil: On the flip side parsers are fun and easy (and can also include regex). :D

charypar: Give up trying to validate an email based on format. The standard is mad and you'll disqualify some valid ones no matter what you do. "Looks like an email" is the best you can do. (@lpil's original suggestion)

services/ses/index.js
AndrewBestbier: Yep

PR #204, opened 2016-11-29
Join Us Section

JoeStanton:
Join Us migration from `website-next` to `website-honestly`.

Depends on #226

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,6 +8,7 @@ dependencies:
   override:
     - npm i -g yarn
     - yarn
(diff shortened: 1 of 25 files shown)


site/pages/join-us/index.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export default function JoinUs({ jobs }) {
```
lpil: Use functional components unless extra stateful features are needed :)

PR #211, opened 2016-12-01
Expose Publish Lambda

JoeStanton:
### Motivation

Exposing the publish lambda as a HTTP endpoint. This endpoint is protected by an API key to prevent unauthorized usage.

This means we can trigger a rebuild from BadgerBot, Prismic Webhooks etc.

### Test plan

- Visit the secret `PUBLISH_ENDPOINT` to trigger a deployment. Confirm a deployment was indeed triggered.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,11 +59,9 @@ services-deploy: dist/services.zip ## Upload the publish service to AWS Lambda
 	$(LOAD_ENV) \
 	&& $(SERVERLESS) deploy
 
-
 publish-service-invoke: ## Invoke the publish service
 	$(LOAD_ENV) \
-	&& $(SERVERLESS) invoke --function publish
-
+	&& curl --fail $$PUBLISH_ENDPOINT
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -59,11 +59,9 @@ services-deploy: dist/services.zip ## Upload the publish service to AWS Lambda
 	$(LOAD_ENV) \
 	&& $(SERVERLESS) deploy
 
-
 publish-service-invoke: ## Invoke the publish service
 	$(LOAD_ENV) \
-	&& $(SERVERLESS) invoke --function publish
-
+	&& curl -XPOST --fail $$PUBLISH_ENDPOINT
```
lpil: dolla dolla?

JoeStanton: Yep - want this variable to be evaluated by bash and not the makefile itself. So it can use the variables set in `LOAD_ENV`.

I'm not very good at Make. I think this is right?

PR #212, opened 2016-12-01
Import a snapshot of website-next

JoeStanton:
During migration, we are planning to reuse the assets/components from `website-next` so we can deprecate the actual `website-next` infrastructure.

After the migration is complete, we'll look to extract and improve the specific components we're using.

This is just a one-off snapshot import, so we can get some sensible diffs for the rest of the incoming features.


JoeStanton:
Need to disable linting and testing of `website-next`

PR #216, opened 2016-12-05
blog featured posts

lazydevorg:
### Motivation

Blog featured posts fetcher


.babelrc
```diff
@@ -6,6 +6,11 @@
   "plugins": [
     "transform-object-assign",
     "transform-class-properties",
-    "babel-plugin-transform-object-rest-spread"
+    ["transform-runtime", {
+      "polyfill": true,
+      "regenerator": true
+    }],
+    "transform-async-to-generator",
+    "babel-plugin-transform-object-rest-spread",
```
JoeStanton: Is there a nicer way to do this now with targets: node or something?

dev/content-fetcher/squarespace_blog.js
```diff
@@ -0,0 +1,30 @@
+export default class BlogFetcher {
+  constructor(fetch) {
+    this.fetch = fetch;
+  }
+
+  getFeaturedPosts() {
+    return this.getPosts({tag: "featured"});
```
JoeStanton: This probably violates some linting rules, do you have ESLint in your editor?

package.json
```diff
@@ -88,5 +88,10 @@
         "make test"
       ]
     }
+  },
+  "devDependencies": {
+    "babel-plugin-transform-async-to-generator": "^6.16.0",
+    "babel-plugin-transform-runtime": "^6.15.0",
+    "babel-polyfill": "^6.16.0"
```
JoeStanton: does the compilation run in development mode?

.babelrc
JoeStanton: Is there a nicer way to do this now with targets: node6 or something?

dev/content-fetcher/squarespace_blog.js
```diff
@@ -0,0 +1,30 @@
+export default class BlogFetcher {
+  constructor(fetch) {
```
JoeStanton: If you're passing in `fetch` to improve testability, you can also utilise a default parameter `fetch = nodeFetch` so we don't have to pass `node-fetch` in app code, but it can be easily swapped during testing?

state/index.js
```diff
@@ -1,9 +1,17 @@
+import fetch from 'node-fetch';
+import Promise from 'bluebird';
```
tabazevedo: Promise should be native to Node 6+ (LTS release) is there a reason we don't use native promises?

tabazevedo: Since we pulled in async/await transpiling, we can do something like:

```
const getSiteState = async () => ({
  featured_blog_posts: await blogFetcher.getFeaturedPosts(),
  ...state
});
```

Very sleek!

state/index.js
```diff
@@ -1,9 +1,12 @@
+import { getFeaturedPosts } from '../dev/content-fetcher/squarespace_blog';
```
JoeStanton: `squarespace-blog`, hyphens not underscores

state/index.js
```diff
@@ -1,9 +1,12 @@
...
+
 const state = {
   contactUsURL: process.env.CONTACT_US_SERVICE_URL,
 };
 
-const getSiteState = () => {
-  return new Promise(resolve => resolve(state));
-};
+const getSiteState = async () => ({
+  featured_blog_posts: await getFeaturedPosts(),
```
tabazevedo: Yeah i found it quite odd. Will make a rule around this.

JoeStanton:
Superseded by #235

PR #217, opened 2016-12-05
Adds LOAD_NODE_ENV variable and BABEL_ENV=node to Makefile

Smeb:
### Motivation
Changes the makefile to use LOAD_NODE_ENV where the environment of the rule uses node.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ SHELL:=/bin/bash
 
 BIN=./bin
 LOAD_ENV=source bin/load-dotenv.sh && source bin/construct-additional-env.sh
+LOAD_NODE_ENV=export BABEL_ENV=node && $(LOAD_ENV)
 NBIN=./node_modules/.bin
 WEBPACK=$(NBIN)/webpack --bail
 MOCHA=$(NBIN)/mocha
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -60,8 +60,8 @@ lint: ## Lint Javascript files
 
 
 services-deploy: dist/services.zip ## Upload the publish service to AWS Lambda
-	$(LOAD_ENV) \
-	&& $(SERVERLESS) deploy
+	$(LOAD_NODE_ENV) \
+  && $(SERVERLESS) deploy
```
lpil: Tabs replaced with spaces here

Makefile
```diff
@@ -33,9 +34,8 @@ dev: badger dist/sw.js ## Run the frontend dev server
 sw: dist/sw.js  ## Compile service worker
 
 fetch:
-	export BABEL_ENV=node && \
-	$(LOAD_ENV) && \
-	node dev/content-fetcher
+	$(LOAD_NODE_ENV) \
+  && node dev/content-fetcher
```
lpil: Tabs replaced with spaces here

PR #229, opened 2016-12-07
Rehydrate State on the Client

JoeStanton:
We need to provide access to the state object used to render the site if we'd like React to take over client-side rendering.

At the moment this places the entire state blob on every page. Obviously, this isn't ideal, but I plan to thread this object through something similar to `stateToProps` for each relevant route to reduce the props to only those relevant for each route.

As a simple first pass though, this solution will unblock us.

### DIFF
--- a/site/index.ejs
+++ b/site/index.ejs
@@ -43,14 +43,11 @@ ga('create', 'UA-16654919-1', 'auto');
 ga('send', 'pageview');
     </script><% } %>
   </head>
-  <% if (typeof state !== 'undefined') { %>
-  <body data-state="<%= state %>">
-  <% } else { %>
-  <body data-state="{}">
-  <% } %>
+  <body>
     <div class="js-app">
       <% if (typeof bodyContent !== 'undefined') { %><%= bodyContent %><% } %>
     </div>
+    <input id="state" type="hidden" value="<%= typeof state !== 'undefined' ? state : '{}' %>">
     <% if (typeof jsPath !== 'undefined') { %><script type="text/javascript" src="<%= jsPath %>"></script><% } %>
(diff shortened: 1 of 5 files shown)


site/index.ejs
```diff
@@ -47,6 +47,11 @@ ga('send', 'pageview');
     <div class="js-app">
       <% if (typeof bodyContent !== 'undefined') { %><%= bodyContent %><% } %>
     </div>
+    <% if (typeof state !== 'undefined') { %>
+      <script type="text/javascript">
```
tabazevedo: We might need to escape the state blob here considering squarespace could send raw html, can recommend [ent](https://www.npmjs.com/package/ent).
```
import decode from 'ent/decode';
import encode from 'ent/encode';

encode(JSON.stringify(state)); // server/compile
JSON.parse(decode(state)); // client
```

grahammendick: I think this is enough to make it safe.

    JSON.stringify(state).replace(/<\/script/g, '<\\/script').replace(/<!--/g, '<\\!--');

Need to make sure that script tags are escaped, https://medium.com/node-security/the-most-common-xss-vulnerability-in-react-js-applications-2bdffbcc1fa0#.n0rfap4jv

site/index.ejs
```diff
@@ -43,7 +43,11 @@ ga('create', 'UA-16654919-1', 'auto');
 ga('send', 'pageview');
     </script><% } %>
   </head>
-  <body>
+  <% if (typeof state !== 'undefined') { %>
+  <body data-state="<%= state %>">
+  <% } else { %>
+  <body data-state="{}">
+  <% } %>
```
JoeStanton: so expressive 😉

tabazevedo:
:+1:

PR #236, opened 2016-12-09
Updates README to indicate need to install deps for website-next

Smeb:
### Motivation
Make the install process easier by adding missing keys and instructions. Removes the dependency on Ruby and instead uses the `http-server` package for hosting static content.


README.md
```diff
@@ -35,6 +35,9 @@ make services-invoke-publish
 npm install --global yarn
 # Install the deps
 yarn
+# Install the deps for website-next
+cd website-next
+yarn
```
Smeb: We could also publish website-next and name it as a dependency, or just have it be ignored by Mocha.

Since we're intending to remove it anyway this works for now.

JoeStanton: "works for now" is fine by me for `website-next` stuff

README.md
```diff
@@ -31,10 +31,14 @@ make services-invoke-publish
 ## Dev setup
 
 ```sh
+# Make sure Ruby is installed
```
JoeStanton: Why?

Smeb: When you run make dev-static it invokes `ruby -run -ehttpd ./dist/static-site/` to server the site. So it's needed for the dev setup.

lpil: Could replace that with a node server but I didn't know how to do it in one line off the top of my head

tabazevedo: Can we replace with a super simple express server?

Smeb: You can add http-server as a dependency and then invoke it.

JoeStanton: `python -m SimpleHTTPServer`

Smeb: You still require python in that instance though so it's the same as using Ruby?

PR #241, opened 2016-12-09
Changes node version to 4.3.0 and adds caching for Yarn

Smeb:
### Motivation
CircleCI is having some issues with caching node v4.3.1. Since v4.3.0 is preinstalled on their Ubuntu / Debian images, we can downgrade to v4.3.0 and save 3 minutes (also some builds were failing due to downloads failing).

Relevant list of supported versions: https://circleci.com/docs/build-image-trusty/#nodejs

We can also start caching Yarn to strip another 1 - 2 minutes off build time.

Relevant instructions to install Yarn: https://circleci.com/docs/install-and-use-yarn/

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,12 +1,21 @@
 machine:
   node:
-    version: 4.3.1
+    version: 4.3.0
   environment:
+    YARN_VERSION: 0.17.8
+    PATH: "${PATH}:${HOME}/.yarn/bin"
     ENVIRONMENT_NAME: staging
 
 dependencies:
+  pre:
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
+  cache_directories:
+    - ~/.yarn
+    - ~/.yarn-cache
   override:
-    - npm i -g yarn
     - yarn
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -1,12 +1,21 @@
...
+    YARN_VERSION: 0.17.8
+    PATH: "${PATH}:${HOME}/.yarn/bin"
     ENVIRONMENT_NAME: staging
 
 dependencies:
+  pre:
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
```
JoeStanton: is Yarn ever preinstalled on CircleCI?

JoeStanton:
Approved, we should move to `4.3.2` when CircleCI resolve their caching issues.

circle.yml
Smeb: From reading their docs I don't think it is. This should cache yarn's own directory (`~/.yarn~), and the package directory (`~/.yarn-cache`).

From looking through the list of preinstallations, node v4.2.0 should be preinstalled, or something like it. CircleCI has:
`circleci-nodejs-4.1.21` and `circleci-nodejs-4.2.61` listed. I can't find exactly what these are, but they look like mappings from the preinstalled major versions to minor versions. The build delay we experienced might be due to some error with these.

Issue #274, opened 2016-12-16
Contact Us nav link currently links to the top of homepage


asavin:
Should instead scroll the homepage to the contact us section.

sarigriffiths:
This looks like solved? @samsmith1983

samsmith1983:
Ah yes, it was fixed bit forgot to close this issue.

PR #276, opened 2016-12-16
Contact us anchor

JRGranell:
### Motivation

Fixed navigating to contact us in the mobile nav menu https://github.com/redbadger/website-honestly/issues/274

### Test plan

Navigate around the site and check that whenever you click the Contact Us link you’re scrolled to the Contact Us section on the home page. Needs testing with and without JavaScript enabled. With JavaScript enabled the ‘Hi Cain,’ textarea should additionally receive the focus.

Although the Contact Us link is only accessible from the small Menu, this should be tested on Desktop as well as Mobile. To test on Desktop, shrink the screen until the small Menu link is visible.

There’s a known issue that when JavaScript is disabled and the user clicks the Contact Us link from the Home Page then the Menu doesn't close automatically

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [x] Windows 7 IE 11
  - [x] Windows 10 Edge
  - [x] iOS 9 Safari
  - [x] Android 6 Chrome
  - [x] Listen to the site on a screen reader
- [ ] Tester approved


site/components/header/small-screen-nav.js
```diff
@@ -29,6 +29,9 @@ export default class SmallScreenNav extends React.Component {
   render() {
     const { navOpen } = this.state;
     const navTabIndex = navOpen ? 0 : -1;
+    const { stateNavigator } = this.context;
+    const contactUsLink = `${stateNavigator.historyManager.getHref(stateNavigator.getNavigationLink('homePage'))}#ContactUs`;
```
JoeStanton: Can you use `scrollIntoView` instead or does that not work very well X-Browser?

JRGranell: @grahammendick Do you want to take that question from @JoeStanton. I think this is not possible if the user is on another page (e.g. about-us).

grahammendick: @JoeStanton We went for a solution that worked without JavaScript. We did investigate putting something in if JavaScript was on, to avoid the full page reload, but it was getting a bit messy. I'm happy to take another look

grahammendick:
On this branch there are two different solutions.
# Works Ok with JavaScript off, but not well with JavaScript on
It appends a hash to the Url and does a full page refresh which focuses to the contact us. It doesn't work well with JavaScript on because it does a full page refresh when there's no need (no other links do full page refreshes).
It still doesn't work great with JavaScript off because if clicking the contact us link from the home page then the small menu nav doesn't close. **_Also, what's the point in making it work well with JavaScript off when the contact us doesn't work with JavaScript off anyway?! Isn't it better to make it work excellently for the JavaScript on case (see below)?_**
The Navigation router doesn't work that well with hashes in the Url so don't see a way to enhance this for the JavaScript on case).

# Works great with JavaScript on, but not at all with JavaScript off
With JavaScript on we can focus the text area the message when the link's clicked. Also we can handle the weird cases like clicking the contact us link when the 'Thank you' message is displayed - we can show the contact us and then focus the text area

grahammendick:
I've had an idea. Forget about hash, it doesn't work well with or without JavaScript (see above). Let's use `autofocus` instead, the HTML5 one that works without JavaScript.

Unfortunately, React strips this attribute out because it has it's own autofocus mechanism. But this mechanism doesn't work without JavaScript so it shouldn't strip it out on the server renderted content (see https://github.com/facebook/react/issues/3066).

To work around we can use the `is` prop to prevent it being stripped out (although a warning is written to console in dev mode)
```jsx
<textarea is autofocus={true} />
```

grahammendick:
The autofocus idea isn't going to work because the server rendered content is statically generated. So can't have different content based on query string parameter, unless we generate two different pages with two different urls, but don't like that idea

site/slices/contact-us-slice/form/index.js
```diff
@@ -15,6 +15,23 @@ class Form extends Component {
...
 
+  componentDidMount = () => {
+    const { stateNavigator } = this.context;
+    stateNavigator.onNavigate(this.focusText);
+  }
+
+  componentWillUnmount = () => {
+    const { stateNavigator } = this.context;
+    stateNavigator.offNavigate(this.focusText);
+  }
```
JoeStanton: We could possibly thread any query string params through to the component automatically here:
https://github.com/redbadger/website-honestly/blob/05f68187534b2ebbc9b3f106e148330d987c2349/site/client/index.js#L27

this would be easier to test and might be more robust?

grahammendick: The `stateNavigator` is put in the context by the Layout component. Should I pass it as prop to the components?

JoeStanton: Hmm it'd be nice if you didn't need to know about `stateNavigator` from this component at all, but now you're not just depending on the query string it's probably not realistic to do another way. Ignore me.

grahammendick: No problem, thanks for the pointer. I've changed things around and removed the `stateNavigator` from the components.

grahammendick:
# Summary of the changes
## Approach
The Contact Us link stores a `contactUs` indicator in the hash of the URL. Because we're using static site generation, this is the only way the scrolling can work without JavaScript. To provide a better user experience with JavaScript enabled, as well as scrolling the section into view, we're setting the focus to the Contact Us textarea so they can start typing straight away (plus, setting focus is good for accessibility).

## StateNavigator workaround
The `StateNavigator` router doesn't currently support hashes in the URL. So, to workaround this issue we've implemented our own `HistoryManager`. You can see this in the `handleContactUsHash` function in `routes\index.js`. The `HistoryManager` is responsible for converting URLs to Hrefs and back again. That means we can pass the `contactUs` indicator as a regular URL query string parameter internally and only convert it to and from a Hash when building anchor elements and browser locations.

site/client/index.js
```diff
@@ -1,33 +1,36 @@
...
-      ReactDOM.render(component, element);
-    };
-    stateNavigator.states[route.key].navigated = render;
+  stateNavigator.onNavigate((oldRoute, route, params) => {
+    const props = route.stateToProps && route.stateToProps(state, params);
+    const pageTitle = typeof route.title === 'function' ? route.title(props) : route.title;
+    const title = `${pageTitle} | ${TITLE_SUFFIX}`;
+    stateNavigator.stateContext.title = title;
+    const component = route.component({ stateNavigator, title }, props);
+    ReactDOM.render(component, element, () => {
```
JoeStanton: Is this callback function going to find a better home?

grahammendick: No probs, I'll create a separate function.

JoeStanton: Is this not the same as doing it within `componentDidMount` in the actual component though?

grahammendick: It's not the same as `componentDidMount` because sometimes the component is already mounted. Clicking the Contact Us link when on the Home page doesn't mount the component but we still need to focus the section

grahammendick: The `afterRender` static function doesn't work that well because the component is the `HomePage` not `ContactUs`. I'll move it into a separate function in `site\client\index.js` and see what you think

samsmith1983:
Found this bug, which we've decided will be fixed separately to this PR: 

Contact us link on mobile nav doesn't navigate to contact us section when coming from blog page #295

samsmith1983:
Looking good!

PR #283, opened 2016-12-19
Change favicon declaration order

### LINKED ISSUE #233: [Blog] Chrome fetches 3 different favicons
home___red_badger

As reported by @StuartHarris 

Should probably be reduced to one?

Etiene:
### Motivation

* Apparently this fixes #233

### Test plan

- Google chrome should fetch only one favicon
- Firefox favicon should remain intact and not change to low res icon

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [x] Android 6 Chrome
- [ ] Tester approved


Etiene:
While this fixes the issue described, I'm not really satisfied because I have nothing support "why". Firefox always grabs last declared favicon but I still didn't find much information on how google chrome should operate. Different sources say it will grab apple-touch if a favicon big enough isn't provided, but when I test this doesn't seem to be true. I also didn't find information on what's causing chrome to make multiple requests for different sized favicons instead of just picking one.

asavin:
This does indeed fixes the original issue!

samsmith1983:
This has caused the incorrect favicon to load on MS Edge 14. Whilst I'm not totally sure which one is now being loaded (can't see it in network tab of dev tools - perhaps I'm looking in the wrong place), the icon previously had nice transparency and looked a correct size. Now it has a white background and is slightly smaller:

screen shot 2016-12-20 at 12 05 19

Etiene:
@samsmith1983 Ok, double check now? :)

Etiene:
@JoeStanton @samsmith1983 Concerning the same issue for the blog template, my local version of the blog ran today (without making any modifications (???)) but I don't get the same issue locally because in fact none of the icons are loading... Wondering if I should just make a PR with this same order of favicons there and see what happens, but the blog repo does not have the same "view deployment" feature. What do you suggest?

samsmith1983:
@Etiene I'd be happy with that

PR #287, opened 2016-12-19
About us timeline

JRGranell:
### Motivation

Timeline slice of the about us page showing text and images for the years 2010 - 2016.

### Test plan

View the about us page, the timeline should render as expected. On IE11 the years do not appear about the timeline which is not intended, but this is how it renders on the mobile version so there is potentially no loss to user experience.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [x] Windows 7 IE 11
  - [x] Windows 10 Edge
  - [x] iOS 9 Safari
  - [x] Android 6 Chrome
- [ ] Tester approved


.flowconfig
```diff
@@ -1,6 +1,8 @@
 [ignore]
 .*/node_modules/.*/test/.*\.json$
 .*/node_modules/fbjs/.*
+.*/node_modules/react-native/.*
```
JoeStanton: What is using `react-native` ?

JRGranell: React-swipeable-views, but it only gets pulled in if you import the native version. Not ideal being on the build server but it doesn't actually get used or pulled into the web version of this package.

JoeStanton: Fair enough

site/components/clientOnly/index.js
```diff
@@ -0,0 +1,29 @@
+import React, { PropTypes, Component } from 'react';
+
+/** Only render the children of this component if JavaScript is enabled and we are in the browser */
+class ClientOnly extends Component {
+
+  componentWillMount() {
+    this.setState({
+      javaScriptSupported: false,
+    });
```
JoeStanton: Can this not just be `state = { javaScriptSupported: false }` as a class property?

JRGranell: I'll update to change what you've suggested

samsmith1983:
Small issue: On Firefox, the years don't go red on hover and are not clickable.

samsmith1983:
Also, at mobile size on firefox, the blobs are not clickable.

samsmith1983:
A couple of known issues on IE11:

- Years don’t appear
- Dots disappear in mobile size

samsmith1983:
The latest change makes the years clickable in firefix, and also fixes the mobile firefox blob issue. However, we now get this in IE11:

screen shot 2016-12-22 at 13 46 54

samsmith1983:
Most recent commit fixes the dodgy years IE11 issue, and firefox buttons are now clickable in desktop and mobile sizes. 

Last remaining issue is that years aren't going red on hover on firefox and IE11.

Issue #298, opened 2016-12-21
Optimise state.json


grahammendick:
The `jobs` and `events` lists are both repeated in `state.json`. They're stored once as an array and again as a keyed object. This eases the generation of the statically generated site, but increases the size of the HTML sent to the browser (the state is sent in a hidden input field).

We should do what we can to streamline the size of HTML. So the `jobs` and `events` list should only be held once in the site state and the work to find the event and job by key put into the static compiler instead.

grahammendick:
Closing this because I don't think it's worth changing at this stage. Although `state.json` is ~twice as big as it needs to be, it's still very small overall

PR #305, opened 2017-01-03
People Listing CMS

grahammendick:
### Motivation

Allows Badgers to be maintained in Prismic.
The following outlines the changes in this and the badger brain Pull Requests.

## Prismic
* Created a new Badger type in Prismic on Staging. The field details are taken fro the [Badger Profile Fields doc](https://docs.google.com/document/d/1NllUgMelb2XQ0yAmB88xCrvzowBGwfu4HT61R0D1E4I/edit?ts=585a8e5c)
* Used the tags field in Prismic to capture the categories for filtering the Badgers.

## Badger Brain
* This PR goes with the PR on badger brain, https://github.com/redbadger/badger-brain/pull/87
* All the Badger fields on Prismic are available on the GraphQL Api even though the website doesn't need them all

## Website Honestly
* The list of categories isn't hard-coded. Instead it's created from the categories assigned to the Badgers. This means there won't be any blank categories in the UI. It also means that any category a Badger is tagged with will appear in the UI so Prismic editors must be careful when tagging Badgers.
* The list of categories deduced from the Badgers returned is added to site state

### Test plan

Unit test the Badger type in badger brain
Unit test the site compile in website honestly
There's no UI so no browser testing required

#### To test on staging:
Check you can create/update/delete badgers in Prismic, following these [guidelines](http://webdocs.red-badger.com/updating-badgers.html)
These badgers won’t appear in the website because the website UI ticket hasn’t been done.
But you can test that these badgers are available through GraphQL:
* Log into Prismic on staging
* Check you can access all badgers from Prismic. Paste the following query into the left hand side:
```
{
  allBadgers {
    id
    firstName
    lastName
    jobTitle
    startDate
    imageUrl
    about
    skills
    influence
    achievements
    linkedin
    twitter
    github    
    categories {
      name
      slug
    }
  }
}
```
* Click the play button and check that the results are returned on the right
* Check you can access a badger by Id from Prismic. Paste the following query into the left hand side, replacing the id with one of the ids returned from the previous query for all badgers:
```
{
  badger (id: "someid") {
    firstName
    lastName
    jobTitle
    startDate
    imageUrl
    about
    skills
    influence
    achievements
    linkedin
    twitter
    github    
    categories {
      name
      slug
    }
  } 
}
```
* Click the play button and check that the right badger is returned on the right

### Pre-merge checklist

- [ ] Documentation
- [x] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


grahammendick:
The Circle CI is erroring on publish. Suspect that's because the corresponding [Badger Brain PR](https://github.com/redbadger/badger-brain/pull/87) hasn't been merged. It's looking at the old GraphQL schema without the new Badger type, so the website honestly badger brain fetcher falls over when it queries for Badgers

site/compiler/test.js
```diff
@@ -133,14 +149,324 @@ describe('site/compiler', () => {
...
+      expect(pages[11].body).to.match(/engineering/);
+      expect(pages[12].path).to.equal('about-us/people/engineering/2/index.html');
+      expect(pages[12].body).to.match(/engineering/);
+      expect(pages[13].path).to.equal('about-us/people/engineering/3/index.html');
+      expect(pages[13].body).to.match(/engineering/);
+      expect(pages[14].path).to.equal('about-us/people/leadership/index.html');
+      expect(pages[14].body).to.match(/leadership/);
+      expect(pages[15].path).to.equal('about-us/people/leadership/2/index.html');
+      expect(pages[15].body).to.match(/leadership/);
+    });
```
JoeStanton: We might want to consider moving these specs to a different file, namespaced after the section of the site they actually check. It might become huge otherwise.

site/routes/definitions.js
```diff
@@ -1,14 +1,27 @@
...
+const createPages = (category, count) => (
+  Array.from(Array(Math.ceil(count / 20)).keys()).map(page => ({ category, page: page + 1 }))
+);
+
+const genBadgersParams = state => (
+  state.categories.reduce((params, category) => {
+    const count = getBadgersByCategory(state.badgers, category).length;
+    return params.concat(createPages(category.toLowerCase(), count));
+  }, createPages('everyone', state.badgers.length + 1))
+);
```
JoeStanton: I think this logic belongs somewhere else - could it go in the fetcher? or a different file in the `state/` folder providing convenience functions for accessing that particular part of state?

site/routes/definitions.js
```diff
@@ -56,6 +69,16 @@ export const routeDefinitions : Array<RouteDefinition> = [
     gen: state => state.events.map(({ startDateTime: { date, month, year }, slug }) => ({ date, month, year, slug })),
   },
   {
+    title: ({ category }) => 'Meet our team' + (category !== 'everyone' ? ` (${category})` : ''),
+    key: 'badgers',
+    route: 'about-us/people/{category?}/{page?}',
+    defaults: { category: 'everyone', page: 1 },
+    urlEncode: (_, key, val) => (val === 'ux & design' ? 'ux-design' : encodeURIComponent(val)),
+    urlDecode: (_, key, val) => (val === 'ux-design' ? 'ux & design' : decodeURIComponent(val)),
```
JoeStanton: Could we handle this in the graphql layer? I.e. produce a slug for categories there?

state/index.js
```diff
@@ -13,16 +13,29 @@ const toDict = (array, keyFn) => array.reduce((obj, item) => ({
...
 
+const getCategories = badgers => (
+  Object.keys(badgers
+    .reduce((uniqueCategories, badger) => (
+      badger.categories
+        .reduce((categories, category) => (
+          categories[category] ? categories : { ...categories, [category]: 1 }
+        ), uniqueCategories)
+    ), {}))
+);
```
JoeStanton: Why don't you move this back into `badger-brain` under the `getBadgers` function? I.e. revert back to having specific functions there so you don't need to postprocess here. `toDict` is slightly different as it's a bit more generic.

JoeStanton: Ah I think you're doing this here to avoid making two requests. Could still move this into `getData`

site/compiler/test.js
```diff
@@ -133,14 +149,324 @@ describe('site/compiler', () => {
...
+        event: {},
+        badgers: [{
+          firstName: 'Alex',
+          lastName: 'Savin',
+          categories: [
+            'Engineering',
+            'Leadership',
+          ],
+        }],
+        categories: ['Engineering', 'Leadership'],
```
JoeStanton: We could actually use my library to go from flowTypes -> test data here? https://github.com/JoeStanton/babel-transform-flow-introspection/blob/master/use-cases/generative-testing.js

JoeStanton: Alternatively, we could just move all of these stubs to `test/state.json` and keep that up to date.

site/compiler/test.js
```diff
@@ -133,14 +149,324 @@ describe('site/compiler', () => {
...
+
+    describe('with one UX & Designer', () => {
+      it('should render one everyone and one ux-design', () => {
+        const pages = compileSite({
+          jobs: [],
+          job: {},
+          contactUsURL: '',
+          featuredBlogPosts: [],
+          events: [],
+          event: {},
```
JoeStanton: We should define a `baseState` with all of these empty fields then you could just do:

```
{ ...baseState, badgers: mockBadgers }
```

To keep the test data focused.

site/fetchers/badger-brain.js
```diff
@@ -64,16 +64,27 @@ const fullEventsQuery = `
...
+      }
     }
   `;
 
   return fetch(badgerBrainEndpoint, getRequestOptions(body))
           .then(response => response.json())
-          .then(events => sortEvents(selectValidEvents(events.data.allEvents)));
+          .then(({ data }) => ({
+            events: sortEvents(selectValidEvents(data.allEvents)),
+            badgers: data.allBadgers,
```
JoeStanton: The postprocessing could go here.

site/routes/definitions.js
```diff
@@ -1,14 +1,27 @@
...
-type RouteDefinition = {|
+type RouteDefinition = {
   title: string | (props: Object) => string,
   key: string,
   route: string,
-  defaults?: any,
   stateToProps?: (state: Object, params?: Object) => any,
   gen?: (state: Object) => Array<Object>,
-  render?: (state: Object) => any,
-|}
```
JoeStanton: It's nice to keep the exact object type here if possible, it'll prevent you misspelling a key in the config. All the functions can just be optional.

site/routes/definitions.js
asavin: Yeah this belongs to the resolver in Badger Brain

site/compiler/test.js
grahammendick: Good idea. I've created a new file called `badgers.test.js` at the same level as `test.js` (thought this was preferable to creating a folder called `badgers` with just test.js in)

site/fetchers/badger-brain.js
grahammendick: Agreed. I've moved `getCategories` out of `state.index.js` into here so that it matches `selectValidEvents`

state/index.js
grahammendick: Moved into `badger-brain` fetcher (as you suggested below) so that it matches `selectValidEvents` with events

site/routes/definitions.js
grahammendick: @asavin @JoeStanton We're not hard-coding a list of valid categories. We're just returning badgers from badger brain with their associated categories. Then, on the fetcher side, we're working out the categories by looping through the badgers to generate a unique list of those assigned. This means that any category assigned in Prismic, e.g., 'example category' will appear as a filter in the website - is this desirable? It also makes it difficult to create slugs because the category list is dynamic.

Shall we switch to a hard-coded list of categories and their slugs returned from graphql?

We could still filter this list on the fetcher side so that only those categories with badgers attached actually appear.

site/compiler/test.js
grahammendick: Test data should live with the test. That way it’s clear what’s being tested. It also means that we can safely edit data for a test without worrying about breaking another test that might be sharing that data.

site/compiler/test.js
grahammendick: @JoeStanton Same as my previous comment

> Test data should live with the test. That way it’s clear what’s being tested. It also means that we can safely edit data for a test without worrying about breaking another test that might be sharing that data.

site/routes/definitions.js
grahammendick: @JoeStanton Turning the list of badgers into different pages per category is a UI thing. So it fits better in the `routes` folder than the `state` folder.

I raised a similar issue on the [Events and Jobs dictionary currently stored in State](https://github.com/redbadger/website-honestly/issues/298). The State should just hold arrays and then logic to manipulate the State for the components should sit in the routes folder?

Doesn’t seem right to store the same list in two different ways in State. (I guess I see `state.json` as a denormalised database)

site/routes/definitions.js
JoeStanton: Yeah I prefer the idea of hard coded. Too easy to accidentally create new categories.

site/compiler/test.js
JoeStanton: Point taken above, but I think here it makes it much clearer which data is actually relevant. All `baseState` would be is the empty keys/values needed to satisfy the rest of the (irrelevant) compilation.

site/compiler/test.js
JoeStanton: Nice - should we call it `about-us/people` or just `people` rather than `badgers` to match the URL? 👍

site/routes/definitions.js
```diff
@@ -1,14 +1,27 @@
...
+}
+
+const getBadgersByCategory = (badgers, category) => (
+  badgers.filter(badger => badger.categories.filter(cat => cat === category).length > 0)
```
JoeStanton: `badger.categories.indexOf(category) !== -1` ?

site/routes/definitions.js
JoeStanton: Is there a way of doing `key-value` mappings in Prismic?

grahammendick: @JoeStanton Should we hard-code this list in the badger-brain fetcher in website honestly or should we create a separate request for it in graphql?

site/compiler/test.js
grahammendick: Ok

site/routes/definitions.js
grahammendick: Just chatted with @JoeStanton and we decided:
* Hard-code the list of categories in grapnql layer
* Each badger returns with a list of their categories. Each category has two properties name and slug
```
{
  firstName: ‘Alex’,
  ...
  categories: [
    {name: ‘Engineering’, slug: ‘engineering’},
    {name: ‘UX & Design’, slug: ‘ux-design’}
  ]
}
```
* The filtering of invalid categories should be done within graphql layer. If a badger has an invalid category assigned (not one of the hard-coded list) then this category shouldn’t be returned.

Issue #186, opened 2016-11-11
Mobile browser chrome being coloured red looks alarming and should be changed


asavin:
![screenshot_20161111-101558](https://cloud.githubusercontent.com/assets/487468/20212899/9c0a996e-a7fd-11e6-866c-a38e4b2a27d9.png)

Top of the browser colour should be changed to Badger Black.

PR #334, opened 2017-01-18
#186: Changes status bar for chrome and safari to black

Smeb:
### Motivation
Refer to issue #186; the red colour for the top bar was alarming and has been changed to badger black in site/index.ejs.


samsmith1983:
Looks good!

PR #343, opened 2017-01-20
Badger profile

Etiene:
### Motivation

- Add badger profile page

TODO: write a motivation.

### Test plan

- TODO: Write a test plan.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


Etiene:
Preview for testing: https://www-staging.red-badger.com/2972415/about-us/people/alex-savin

grahammendick:
The unit tests for the compiler need some tweaks:

1. The tests in the `people.test.js` shouldn’t’ve been changed. These tests are for the people listing and should be kept independent of the profile pages. The only thing that would need changing is the `assert` for the `pages.length` because now there’s extra pages generated. 
2. What you should do instead is create separate unit tests for the badger profile pages. If there aren’t many then they should go in the `test.js` file inside the compiler folder.
3. The object you pass into `compileSite` in the tests must match the object in `state.json`. You can’t pass in `badger` if that doesn’t exist in `state.json`. There are two different approaches:
    1. Add badger into `state.json` by changing `state/index.js`. This is what `jobs` and `events` has done but I prefer the second option below
    2. Change the `stateToProps` on the badger route to find the badger from the `badgers` array. I prefer this because it means that badgers are only stored in `state` once. The difference in performance is negligible for the size of lists we’re talking about

Etiene:
@grahammendick 
1&2 - Alright! :) I'm going to move some things, but it's still going to need modifications because the state is different and compile site wouldn't work.
3 - The state has been changed and it works exactly like events and jobs for the moment. I could change to a different approach now, but I did like events and jobs because (1) consistency and (2) they were the examples I had on how to do this.

grahammendick:
I think if you add `badger:[]` to the object passed into `compileSite` of each test in people.test.js then they should be fine?
That's cool to go with the events and jobs approach, but you'll need to change `state/index.js` to match

Etiene:
`state/index.js` is already matching it!

grahammendick:
oops, so it is, sorry

Etiene:
Tried leaving an empty `badger` object at the test state, that will generate an error on `compileSite` because the `stateToProps` on the individual page then breaks. That means most of my changes will have to stay there. I'll move part of it, but I think it could be kept at the same file at the end since both the profile page and the listing are under `about-us/people/X`.  Either way I still need to find out why it's timing out.

grahammendick:
I see what you mean. That's a shame.
Are the tests timing out? I guess they're generating a lot more pages now - in the last test they're generating 62 profile pages!

grahammendick:
Here's an idea. In test, we could turn off the `renderToString` because this is the expensive operation. Then the compiler tests could just assert that the correct folders and files are created without checking the contents.

If we want to assert on the contents then this should go in component level tests. For example, we could move the `people.test` content asserts into `test.js` file inside the `pages/meet-our-team` folder.

If you like the idea, we can chat about how best to turn off the `renderToString` in tests

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,116 @@
...
+          </div>
+          <div className={styles.authorLinks}>
+            {getBlogsLink(badger)}
+          </div>
+          <hr />
+          <a className={styles.greenBox} href="/about-us/people">See Everyone</a>
+          {badger.categories.map(c =>
+            <a className={styles.greenBox} href={'/about-us/people/category/' + c.slug}>See {c.name} team</a>
+          )}
+        </div>
```
grahammendick: The Categories search links should use the `Link` component so that the navigation happens on the client, without going back to the server

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,116 @@
...
+      </li>);
+    }
+  }
+  return socialItems;
+};
+
+const getBlogsLink = badger => {
+  if (badger.squarespaceId) {
+    return (<div className={styles.authorLink}>
+      <a href={'https://red-badger.com/blog/?author=' + badger.squarespaceId}>
```
grahammendick: Don’t hard-code the domain 'https://red-badger.com'. I think it should just be '/blog/?author='

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,116 @@
...
+          <InlineSVG
+            src={s.img}
+            title={s.social}
+          />
+        </a>
+      </li>);
+    }
+  }
+  return socialItems;
+};
```
grahammendick: Instead of functions like `getSocialItems` that return JSX, convert them into Components, e.g., const SocialItems = badger => … and then you can use it like <SocialItems badger={badger} />

grahammendick:
Code looks good! Just a few review comments above

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,77 @@
...
+            {badger.achievements && [
+              <h2 className={styles.subheader}>Achievements at Red Badger</h2>,
+              <p>{badger.achievements}</p>]}
+
+            {badger.influence && [
+              <h2 className={styles.subheader}>The thing that changed me</h2>,
+              <p>{badger.influence}</p>]}
+          </div>
+          <div className={styles.authorLinks}>
+            {getBlogsLink(badger)}
```
grahammendick: You should do the same thing to `getBlogsLink` as you did to `getSocialContacts`. You don't need to put it in a separate file, you can just change it to `const BlogsLink` and then you can use it like `<BlogsLink...`

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,77 @@
...
+          </div>
+          <hr />
+          <a className={styles.greenBox} href="/about-us/people">See Everyone</a>
```
grahammendick: Change this anchor to a `Link` component like you did with the categories

Etiene: ahh woops

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,78 @@
...
+              <p>{badger.influence}</p>]}
+          </div>
+          <div className={styles.authorLinks}>
+            <BlogsLink badger={badger} />
+          </div>
+          <hr />
+          <Link to="badgers" className={styles.greenBox}>See Everyone</Link>
+          {badger.categories.map(c =>
+            <Link
+              to="badgers"
```
grahammendick: You should add a `key` prop to components rendered in a map/loop. See https://facebook.github.io/react/docs/lists-and-keys.html#keys.
The SocialContacts component needs a similar change

site/pages/badger-profile/index.js
```diff
@@ -0,0 +1,79 @@
...
+              <p>{badger.skills.join(', ')}.</p>]}
+
+            {badger.achievements && [
+              <h2 className={styles.subheader}>Achievements at Red Badger</h2>,
+              <p>{badger.achievements}</p>]}
+
+            {badger.influence && [
+              <h2 className={styles.subheader}>The thing that changed me</h2>,
+              <p>{badger.influence}</p>]}
+          </div>
```
grahammendick: React is still complaining about missing `key` prop (have a look in the console and you'll see the warning). Instead of adding a key prop I think you should get rid of the array syntax here because it's not needed

site/pages/badger-profile/social-contacts/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+const SocialContacts = ({ badger }: { badger: Badger }) => {
+  const socialItems = [];
+  const socialPages = [
+    { social: 'Twitter', img: twitterSVG },
+    { social: 'Github', img: githubSVG },
+    { social: 'LinkedIn', img: linkedinSVG },
+  ];
+
+  for (const s of socialPages) {
```
JoeStanton: I think it would be nicer to filter then map/reduce `socialPages` here instead?

site/routes/definitions.js
```diff
@@ -64,10 +64,11 @@ export const routeDefinitions : Array<RouteDefinition> = [
     gen: genBadgersParams,
   },
   {
-    title: 'Badger',
+    title: ({ badger }) => badger.firstName + ' ' + badger.lastName,
```
JoeStanton: You should probably also do `[firstName, lastName].join(" ")` here like you did previously

site/types/badger.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable */
+// @flow
+
+declare type Badger = {
```
JoeStanton: You don't need the `declare` keyword here.

site/routes/definitions.js
Etiene: alright!

site/types/index.js
```diff
@@ -1,5 +1,7 @@
 import type { Tweet } from './tweet';
 import type { InstagramPost } from './instagramPost';
+import type { Badger } from './badger';
 
 export { Tweet };
 export { InstagramPost };
+export { Badger };
```
JoeStanton: `export {Tweet, InstagramPost, Badger}` or we can use the draft support for re-exporting: http://jamesknelson.com/re-exporting-es6-modules/.

site/types/badger.js
Etiene: I was following consistency with other types in this folder. Should I refactor others?

site/types/index.js
Etiene: alright!

site/pages/badger-profile/social-contacts/index.js
Etiene: Hm. Ok, let me try that :)

site/pages/badger-profile/index.js
Etiene: It was complaining about missing enclosing parent before. Ok, I guess that is preferable then  :)

site/pages/badger-profile/social-contacts/index.js
Etiene: @JoeStanton please take a look? Is this what you meant? :)

site/pages/meet-our-team/test.js
```diff
@@ -0,0 +1,152 @@
...
+      expect(badgersList).to.match(/Alex 18/);
+      expect(badgersList).to.match(/Alex 19/);
+      expect(badgersList).to.not.match(/Are you a potential Badger/);
+      expect(previousLink.attr('href')).to.equal(undefined);
+      expect(previousLink.text()).to.equal('Previous page');
+      expect(nextLink.attr('href')).to.equal(undefined);
+      expect(nextLink.text()).to.equal('Next page');
+    });
+  });
+});
```
JoeStanton: I think some of these tests are starting to look a bit over-specified. Might be worth a chat to decide if this is too detailed. Fine for this PR though!

Etiene: okay! :)

site/index.ejs
```diff
@@ -42,7 +42,7 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 ga('create', 'UA-16654919-1', 'auto');
 ga('send', 'pageview');
     </script><% } %>
-    <link rel="stylesheet" type="text/css" href="https://cloud.typography.com/7838134/7848772/css/fonts.css">
+    <link rel="stylesheet" type="text/css" href="https://cloud.typography.com/7838134/6842972/css/fonts.css" />
```
JoeStanton: Does this need to change on the Squarespace template too?

Etiene: Not yet. This is introducing a new font pack, which is breaking in Safari (see this, @samsmith1983), and we will have to repack things again before going to live. @asavin understands these details better than me to be able to explain.

samsmith1983:
The profile image is not scaled / positioned correctly on iPhone 5S (running iOS9):

image1

samsmith1983:
The `See Everyone` etc boxes are too tall compared to the design.

Design:
screen shot 2017-01-26 at 12 33 37

Current:
screen shot 2017-01-26 at 12 33 30

samsmith1983:
On mobile, the text size of the persons name is too small. In the design it's 40px, and it's currently 30px:

Design:
screen shot 2017-01-26 at 12 36 52

Actual:
screen shot 2017-01-26 at 12 35 50

samsmith1983:
The font is slightly too small on tablet size too.

samsmith1983:
Links to twitter, github and linkedin should probably open in a new tab (is that what we normally do as a rule?)

samsmith1983:
^ Actually, it looks like we don't currently open new tabs elsewhere on the site, so ignore this for now.

samsmith1983:
@Etiene This is is pretty much smallest thing in the world, but could you change the colour of the divider below `Read X's blog posts` to make it match the design (should be lighter than it is at the moment).

Current:
screen shot 2017-01-27 at 13 57 05

Design:
screen shot 2017-01-27 at 13 57 14

Etiene:
@samsmith1983 Does it look ok now? (wait for CI)

samsmith1983:
Looks beautiful!

Another tiny thing... According to the design, `Signature Skills` should actually be `Signature skills` (with a small S).

@Etiene

samsmith1983:
💥

Etiene:
Yayy 🚀

samsmith1983:
@Etiene Actually just spotted one more thing. The twitter, github and linkedin links should have better accessibility titles. Currently they are things like:

screen shot 2017-01-27 at 15 59 08

Best way to repro:

- Cmd+F5 to enable voiceover
- Press Ctrl+option+U to get list of links

You'll see the above type code.

Etiene:
alright! will get to this after the brand intro :)

Etiene:
badger badger🚀

PR #355, opened 2017-01-31
Webpack 2

JoeStanton:
### Motivation

Webpack 2 has now been released, it contains a number of nice new features. One feature we're not yet taking advantage of is tree shaking.

Unfortunately, most modules we depend upon are currently using CommonJS module syntax, not ES6 syntax, this means we only save 2.2% on the bundle with tree shaking enabled, this should increase a lot in the future.

### Test plan

- A basic regression test. Do the JS and styles load correctly? Does the build work? (Yes)

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


package.json
```diff
@@ -36,8 +36,8 @@
     "eslint-plugin-import": "^1.15.0",
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.2.1",
-    "extract-text-webpack-plugin": "^1.0.1",
-    "file-loader": "^0.9.0",
+    "extract-text-webpack-plugin": "grrowl/extract-text-webpack-plugin#order-undefined-error",
```
JoeStanton: This fork allows ignoring the undefined order of our CSS modules. We should fix this at its root cause but it's not entirely clear how to do that at the moment: https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/166

grahammendick: The "Order in extracted chunk" errors have decreased, but not gone. There were 44 and there are now 12. They all flag up against the same file `./site/components/header/style.css`

webpack.base.config.js
```diff
@@ -40,34 +40,24 @@ const baseConfig = {
...
     new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
-    new webpack.optimize.DedupePlugin(),
-    new webpack.optimize.OccurenceOrderPlugin(),
     new webpack.EnvironmentPlugin(Object.keys(process.env)),
-    new ExtractTextPlugin(
-      'assets-honestly/styles-[contenthash:base64:5].css',
-      { allChunks: true }
-    ),
+    new webpack.LoaderOptionsPlugin({ options: { context: __dirname, postcss: [autoprefixer] } }),
+    new ExtractTextPlugin({ allChunks: true, filename: 'assets-honestly/styles-[contenthash:base64:5].css', ignoreOrder: true }),
```
grahammendick: This same `ExtractTextPlugin` change needs to go in `webpack.dev.browser.config.js`

package.json
JoeStanton: The point is that they don't fail the build with this fork. I did try and correct the issue in `header/style.css` but didn't make much progress. Practically, it won't matter unless there are styles with exactly the same level of specificity being generated in a non-deterministic order.

webpack.base.config.js
JoeStanton: Ah yes - I won't be able to work on this today though.

asavin:
Closing this one for now as it is stale. Feel free to re-open.

Issue #325, opened 2017-01-09
Open page (background content) shouldn't be screen-readable when mobile nav is open


samsmith1983:
Split this one from #171. We've fixed the scrolling, but the background content is still screen-readable when the mobile nav is open.

Repro:

- Grab an iPhone
- Turn on Voiceover (Settings -> General -> Accessibility)
- Go to red-badger.com
- Click in visible bit of text in background

It's read out to you. It shouldn't be selectable.

grahammendick:
@asavin @dstodolny There are two issues to do with the background content when the small nav is open: It shouldn’t be scrollable and it shouldn’t be (screen) readable.

The fix put in for the first (#171) was to prevent scrolling when the small nav is open. The problem with this is that it prevents the small nav menu from scrolling, so some menu items might be unreachable.

I think there’s a different fix we could try that will fix both the issues together. **When the small navigation window is open we hide the background**. Because the background’s hidden, it won’t be scrollable or (screen) readable. What do you think?

dstodolny:
If it works it's better. My fix is a little bit hacky.

grahammendick:
It doesn’t work. The small nav menu slides out so how we can't hide the background during the animation.

gregdorward:
Still an issue

dstodolny:
This is fixed now

Issue #171, opened 2016-11-04
Open page (background content) shouldn't be scrollable when mobile nav is open


zoeabryant:
- [ ] users shouldn't be able to scroll background content

- [ ] screen reader users shouldn't be able to read background content

sarigriffiths:
Is this same as #197 and #198 but for the Honestly? @samsmith1983 - please fix.

samsmith1983:
I'm not sure this is the same thing (repro steps are a bit scarce!) . 

I think this issue refers to when the mobile nav is open - it's saying that users shouldn't be able to then scroll through the content in the background (ie. the open page), and screen readers shouldn't be able to see the open page.

samsmith1983:
This is mainly fixed in #323. However I've split the screen reader bug into a separate issue, and also logged another improvement issue:

Mobile nav menu should be scrollable #324
Open page (background content) shouldn't be screen-readable when mobile nav is open #325

samsmith1983:
Should have closed this one... The main issue ("Open page (background content) shouldn't be scrollable when mobile nav is open") is now fixed. Remaining issues are logged separately.

PR #380, opened 2017-02-15
He disabled background scrolling with this one simple trick. Senior developers hate him. 😱

dstodolny:
### Motivation

- [Related story](https://github.com/redbadger/website-honestly/issues/325)

😎 Prevents background content to scroll when menu is open. 
😎 The menu itself stays scrollable.

### Test plan

- This only applies to mobile devices while touching the screen.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [x] Windows 7 IE 11
  - [x] Windows 10 Edge
  - [x] iOS 9 Safari
  - [x] Android 6 Chrome
  - [x] Listen to the site on a screen reader
  - [x] Navigate the site using the keyboard only
- [ ] Tester approved

### DIFF
--- a/site/components/header/small-screen-nav.js
+++ b/site/components/header/small-screen-nav.js
@@ -14,37 +14,23 @@ export default class SmallScreenNav extends React.Component {
   }
 
   handleInputChange = event => {
-    this.scrollLock();
+    if (event.target.checked) {
+      document.body.style.position = 'fixed';
+      this.smallScreenNav.scrollTop = 0;
+    } else {
+      document.body.style.position = 'relative';
+    }
     this.setState({
       navOpen: event.target.checked,
     });
   }
 
   closeMenu = () => {
-    this.scrollRelease();
+    document.body.style.position = 'relative';
     this.setState({
       navOpen: false,
     });
     return true;
-  }
-
-  scrollLock = () => {
-    this.smallScreenNav.addEventListener('wheel', this.onScrollHandler, false);
-    this.smallScreenNav.addEventListener('touchmove', this.onScrollHandler, false);
-  }
-
-  scrollRelease = () => {
-    this.smallScreenNav.removeEventListener('wheel', this.onScrollHandler, false);
-    this.smallScreenNav.removeEventListener('touchmove', this.onScrollHandler, false);
-  }
-
-  onScrollHandler = e => {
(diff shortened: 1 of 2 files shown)


asavin:
Bonus points for a catchy title!

samsmith1983:
@dstodolny so is this just a 'better' fix for #171?

#325 is actually to do with screenreaders, not scrolling

asavin:
Preview app https://www-staging.red-badger.com/f089a78/ doesn't seems to fix the issue on mobile?

dstodolny:
Will see why is that. Locally seems to be fine. :(

dstodolny:
Now it should be fine.

samsmith1983:
@dstodolny unfortunately this doesn't seem to be working for me. Tested on iOS9 Safari on iPhone 5S.

It actually seems to have regressed back to how it was before the original fix for #171 - ie. the background scrolls but the menu doesn't.

dstodolny:
Have you tried that on real iphone. I think browserstack treats touch events a little bit funny.

samsmith1983:
@dstodolny that was on a physical iPhone

dstodolny:
I found out that it's not working only on iphone 5. Will try to fix this.

asavin:
Seems working on my phone - happy to approve it if it passes tests.

If you open the menu then rotate the phone to landscape, "Close" link doesn't work anymore.

samsmith1983:
This is now working as expected on all tested devices, including iPhone 5S. The mobile nav menu is now scrollable, and the background is not. 

Alex found a slight edge case (see above), but this is specific to iOS10 Safari on iPhone 6/7 plus, and actually it does work around half the time. Odd - but it's such an edge case we're going to let it slide.

Will merge this pending code review.

dstodolny:
In my opinion it's unrelated bug that could be reported. It happens because iPhone 6/7+ landscape mode passes our mobile breakpoint.

samsmith1983:
@dstodolny need to confirm this with Alex, but I think it's actually a regression - in which case it's not unrelated. But it's a very minor edge case, so fine to ship with. I'll get a minor issue logged.

PR #399, opened 2017-03-06
Website next obliteration

asavin:
giphy 20

It's not a PR we like, but it is a PR we all deserve.

### Motivation

Back in a time we used to have a separate app to handle Events, News and Jobs pages. It was called Website Next, and around the same time last year it was the cutting edge of web app tech. Last December we have merged all components from Website Next into Website Honestly. We did not however properly integrated those components, but left them to rot in their own '/website-next` folder, alone and forgotten.

But no more - this PR brings those components into the Website Honestly folder structure.

It also fixes most of the ESlint issues to maintain our DX sanity.

### Test plan

- Test everything

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


asavin:
On the remaining scope of this PR - I think the legacy typography should be dissolved and we should use main typography styles from the Honestly. Components should probably be reviewed and relocated potentially some to the`/pages` and some into the `/components`.

It's fair to assume that Jobs and Events will be redesigned in a not too distant future (famous last words). Don't think there are any concrete plans though just yet.

site/components/grid/index.js
```diff
@@ -1,3 +1,8 @@
+/*
+* This component is ported from Website Next and is meant to be
+* used only by Jobs and Events components
+*/
+
```
asavin: I'll review once again the components, and might as well get rid of these comments, or refine them somewhat.

grahammendick:
@asavin  Yes, good spot with `icons`. Also `tags-list` folder can be deleted, right?

grahammendick:
Looks good. I think `wall` and `event-title` are only used by one component so probably should move into pages. But more important to merge this, I think

PR #407, opened 2017-03-10
add opengraph tags to meet our team page

dstodolny:
### Motivation

- Added opengraph tags to meet our team page

### Testing

It can only be tested after merging this to live


grahammendick:
@dstodolny Why not create an Open Graph component, e.g., in `meet-our-team/open-graph/index.js` and move the code into there?

grahammendick:
@dstodolny I just noticed that there are other examples that are done the same way as yours. Maybe it's better to keep it like you've done it for consistency. What do you think?

grahammendick:
@dstodolny Another idea is to create a Social component inside the components folder and move the images out of assets into there as well. Then you could refactor the other three uses to all use this Social component. It's more work and would need regression test of the other three places. What do you think?

dstodolny:
We can do that but it cannot be tested before deploying it to live 😢

grahammendick:
@dstodolny no problem, do you think we should refactor it or leave it as it is?

dstodolny:
Refactor because probably we will add opengraph to more pages

grahammendick:
The twitter and open graph images are hard-coded, e.g., https://red-badger.com/assets-honestly/social/meet-our-team-OG.png. This isn’t the way we handle other images. With other images, we import them and they get turned into a unique string by file loader. That means we can update images without worrying about the old ones being cached.

What happens if we want to update the twitter and open graph images? Will twitter ignore the new ones because it’s already got the old one cached? Or will it always fetch the new one when the link is shared. Even if it does fetch the new one, it still might get the cached old one.

I think it’s safer to change the code to import the images like normal. So I think we should refactor all the twitter and open graph code into a single `Social` component into the shared `components` folder. Then we can import the images and pass them as a prop to this shared component.

grahammendick:
[Facebook caches the image](http://stackoverflow.com/a/15682504) so I think we should change the code to import images instead of hard-coding them.

grahammendick:
@gregdorward @chris-brookes We've refactored all the open graph code. There are four places where it's used (see below). So can you please retest all of them when it's in live. 
* Home page
* Our work
* What we do
* Meet out team

PR #425, opened 2017-03-29
make paddings static

dstodolny:
### Motivation

- this reverts case study paddings to be static


chris-brookes:
@dstodolny - Do you also need to remove this? You added it in when you did the original change.

screen shot 2017-03-29 at 16 04 52

dstodolny:
nope, it should remain like that

PR #428, opened 2017-03-31
add open graph tags to the retailer case study page

dstodolny:
### Motivation

- added open graph tags.

### Test plan

- Only on live


gregdorward:
This one is going to need to be tested in production, so will just need a code review to get into Staging

Issue #431, opened 2017-04-05
Adhering to colour contrast accessibility guidelines


gregdorward:
**See last comments for the resolution**

**Description of the issue:**
There are currently a number of places throughout Website Honestly where we are failing to meet accessibility guidelines for sufficient colour contrasting. The guidelines state that when using text under 19 pixels in size, a contrast of 4.5:1 should exist.

With Badger Red (#FC1D42), these are our current contrasts:

#FC1D42 against a background of #F8F8F8 - contrast of 3.67:1
#FC1D42 against a background of #212121 - contrast of 4.14:1

To fix this, we will either need to come up with 2 new types of red to use (one darker, one lighter), or increase the font size wherever this colour is used to 19 pixels

**Browser used:**
Chrome

**OS used:**
OS Sierra

**Device used:**
Macbook

**Steps to reproduce:**

1. Navigate to Website Honestly
2. Run an accessibility check on each page (e.g. WAVE)

**Actual result:**
As described above, we're currently falling short of some accessibility standards

**Expected result:**
Colour contrast of at least 4.5:1 should exist throughout the site

gregdorward:
cc @sarigriffiths & @Roisi

gregdorward:
Below are two potential options for darkening and lightening the reds respectively. These reds look to be just within the range of what's acceptable, but may not be right at the limit of what we can get away with

#e5051a against a background of #f8f8f8

<img width="966" alt="screen shot 2017-04-21 at 09 53 05" src="https://cloud.githubusercontent.com/assets/20296527/25270361/56772bd4-2678-11e7-81d9-3d5f743c750e.png">

#ff316b against a background of #212121

<img width="1008" alt="screen shot 2017-04-21 at 09 51 56" src="https://cloud.githubusercontent.com/assets/20296527/25270336/2e14e500-2678-11e7-9fab-d0f8ced056cf.png">

sarigriffiths:
Happy with the darker red (#e5051a) for the light grey background.
Can we change the lighter red for the dark background to be: #fc3a54? Making it less purple. And let's keep #212121 grey.

Just wondering - why Sentinel font not rendering on those screenshots? They are Georgia ^. @gregdorward

gregdorward:
Not sure why it's not rendering the Sentinel font @sarigriffiths . It could be because these screenshots were taken from within a WAVE accessibility report, so it could be that they render it differently. I'll see if I can find out

gregdorward:
[...]

PR #441, opened 2017-04-20
Q and a slice

grahammendick:
### Motivation

As a potential customer, I want to know Red Badger's answers to common questions, so that I can decide if Red Badger might be right for me.

### Test plan

Check the Q&A slice on the About Us page matches the designs on Zeplin.

### Pre-merge checklist

- [x] Documentation
- [x] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


site/pages/about-us/q-and-a-slice/category/style.css
```diff
@@ -0,0 +1,51 @@
...
+
+@media mediumScreen {
+  .category__title {
+    width: calc((100% + 20px) / 6);
+    padding-right: 20px;
+  }
+
+  .category__questionList {
+    width: calc((100% + 20px) * 5 / 6 - 20px);
+  }
```
mveritym: I like this a lot 🎉 do we have to care about browser back-compatibility? I'm not sure how far back `calc` is supported. I imagine it's fine since we're using `flex` which is new-ish as well.

site/pages/about-us/q-and-a-slice/topic/index.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /* eslint-disable react/no-danger */
+  /* eslint-disable jsx-a11y/no-static-element-interactions */
+  render() {
+    const { slug, question, answer } = this.props;
+    const { open } = this.state;
+    const strongText = `<strong class="${styles.topic__answer__strong}"`;
+    const externalLink = '<a rel="noopener noreferrer" target="_blank"';
+    const formattedAnswer = answer.replace(/<strong/g, strongText).replace(/<a/g, externalLink);
```
mveritym: This is really nice.

site/pages/about-us/q-and-a-slice/category/style.css
grahammendick: Thanks, glad you like it 😄 Calculating everything with %, with tweaks to cater for the gutters, means we can match the zeplin grid/column designs exactly. `calc`'s supported by all the browser we care about (and we've used it in a few other places already)

grahammendick: Just to breakdown that calculation:
A medium screen has six columns. The question list should take up 5 of those six columns. But one of those 5 columns doesn’t include the gutter. Gutters take up 20px on medium screen. So, we add on one gutter so we can take 5 / 6 of that. Then we remove the gutter at the end.

gregdorward:
The red text used for the Question Categories is currently failing accessibility checks for sufficient colour contrasting. There is a ticket already open to address this throughout the site though, seeing as this is Badger branding, so I'm not sure whether we should change it in isolation here.

Here's the ticket https://github.com/redbadger/website-honestly/issues/431

gregdorward:
Requested changes:

- [ ] Bit of an edge case, but if we have a Category Type that's longer than the question set, there's not enough padding applied underneath the Category Type. I'm guessing the padding is usually just applied under the questions and answers?

screen shot 2017-04-20 at 17 13 55

- [ ] The expanding of questions to reveal the answers doesn't work when you have javascript disabled. Would we need to have these expanded by default in these cases?

mveritym:
@gregdorward do you think it makes sense to remove the '-' button to collapse the answer in the non JS case?

gregdorward:
Yeah, I'd say so @mveritym

site/pages/about-us/q-and-a-slice/category/style.css
```diff
@@ -22,7 +22,8 @@
   width: 100%;
 }
 
-.category__questionList li:last-of-type h4 {
+.category__questionList li:last-of-type h4,
+.category__questionList li:last-of-type p {
   padding-bottom: 0;
```
gregdorward: Hey @mveritym , the other h4 elements in the question list have a padding-bottom value of 20px. I think that also needs to be applied to the last question, as it's still aligned differently

mveritym: You're right I missed that thanks.

mveritym: @gregdorward I've changed the way I've done the padding to something much simpler. Let me know if it works or not.

chris-brookes:
@dstodolny - Also some other padding changes to do.
Extra padding above and below the title, and also an increase below the questions.
Increase to:
60px above the title,
40px below the tilte,
70px below the questions.

screen shot 2017-04-26 at 15 13 41

chris-brookes:
@dstodolny - A few more changes for you.

screen shot 2017-04-27 at 12 40 19

screen shot 2017-04-27 at 13 05 39

site/pages/about-us/q-and-a-slice/style.css
```diff
@@ -36,6 +36,8 @@
   .qAndA__wrapper {
     max-width: 1400px;
     margin: auto;
+    padding-left: 10px;
+    padding-right: 10px;
```
gregdorward: Does this need to be 20px on both sides to represent a whole gutter?

chris-brookes:
@dstodolny - You are still having two sets of margin at the bottom that are adding together to make the margin too big.

screen shot 2017-04-27 at 17 28 22

site/pages/about-us/q-and-a-slice/topic/index.js
```diff
@@ -0,0 +1,78 @@
...
+import styles from './style.css';
+
+export type TopicProps = {
+  slug: string,
+  question: string,
+  answer: string,
+}
+
+class Topic extends Component {
+  constructor(props: TopicProps) {
```
romanschejbal: This could be replaced with
```
state = {
  open: false,
  showButton: false,
};
```
(for future reference)

dstodolny: Doesn't work :( Do you have any example of that somewhere?

romanschejbal: I think it needs the babel `transform-class-properties` plugin https://github.com/redbadger/aimia-poc-web/blob/master/containers/data-provider/index.js#L4

gregdorward:
Another 'Don't Merge' flag added, as we're waiting on final content sign off for the Q & As

PR #444, opened 2017-04-21
F&M Tea Case Study

mveritym:
### Motivation

I want to know about expensive tea.

### Test plan

* Check the tea designs match those on Zeplin.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved

catDJ


mveritym:
I need two things before I can complete this:
* The tea post image is a jpg with a white background, it needs to be a png with a transparent background or have the background match the off grey colour of the case study page
* What is the page title and slug for the tea case study? There's no existing tea case study to steal this from :)

chris-brookes:
@mveritym - The main header is not expanding across the whole page when the page is wider than 1400px. We need to add background-color: #ace1e6; to the style__header class.

mveritym:
@chris-brookes I don't think adding a background color will work here because of the way the image cuts off the handle of the tea strainer on the right hand side of the illustration. I can try to zoom the image like in the other F&M case study instead.

chris-brookes:
@mveritym - I asked Sari and she wanted the colour to be spread across the whole page, so she is going to re-create the image and make sure the handle is not cut off, then we can add the background colour.

mveritym:
@chris-brookes perfect. I made a change to see how it would look zoomed, would you mind taking a look just in case we don't have to waste sari's time?

chris-brookes:
@mveritym - I don't think the picture is big enough to do that. It makes the height of the image change in a small desktop view, and so the distance the title cuts into the picture also changes, which is totally different to all the other pages.

chris-brookes:
@mveritym 
The Tell us more contact box is still too narrow and the text does not fit on the button when viewed at 980px width.
This is because it is using the margin-right settings from '@media screen and (min-width: 690px).style__contactBox___'. This setting needs to be changed from 39px to 30px, to match the other case study pages.

site/compiler/test.js
```diff
@@ -18,18 +18,17 @@ describe('site/compiler', () => {
...
 
-      expect(routes.length).to.equal(11);
+      expect(routes.length).to.equal(13);
       expect(routes[0].filePath).to.equal('index.html');
       expect(routes[1].filePath).to.equal('what-we-do/index.html');
       expect(routes[2].filePath).to.equal('our-work/index.html');
       expect(routes[3].filePath).to.equal('about-us/index.html');
       expect(routes[4].filePath).to.equal('about-us/join-us/index.html');
       expect(routes[5].filePath).to.equal('about-us/events/index.html');
-      expect(routes[6].filePath).to.equal('our-work/case-study/retailer/index.html');
```
dstodolny: Why this was removed?

dstodolny: Nevermind :) I see it was moved to another test.

site/css/typography/_fonts.css
```diff
@@ -96,6 +96,11 @@
   line-height: 1.2em;
 }
 
+.fontM3 {
+  font-size: 24px;
```
dstodolny: I think we shouldn't add new typography styles as everything we use should be taken from: http://brand.red-badger.com/d/WEl30uS4YzKw/red-badger-styleguide-style-guide#/elements/typography

mveritym:
@dstodolny I think this is ready for re-review :)

dstodolny:
jackiechangivesathumbsupnumber1

gregdorward:
'Don't Merge' flag added, as this needs internal approval for copy, followed by external (F&M) sign off for copy and design

martinhartt:
Once the Camden Market case study is merged, the `content__title` selector should use `fontM` rather than `fontM2` for consistency with the new case studies.

gregdorward:
Blocked until #557 is approved by F&M

gregdorward:
This still needs final copy, so the 'Do not merge' flag should remain

pataruco:
I am closing this PR because *F&M* because this branch has diverged too far from master and the feature has implemented

Issue #438, opened 2017-04-18
Typography is Blocking


grahammendick:
**Description of the issue:**
See #434 for background. We display web fonts by making a stylesheet request to Typography (which then redirects to the relevant css file on our servers). Be default, stylesheet requests are blocking. The user won’t see any content until the Typography request returns.

**Browser used:** N/A
**OS used:** N/A
**Device used:** N/A

**Steps to reproduce:**
I don't know of an easy way to slow down just the Typography request. Instead I wrote my own test-server to play the role of Typography.com:

1. Create a node express app that redirects to the css file after 5 seconds 
```
var express = require('express')
var app = express()

app.get('/', function (req, res) {
  setTimeout(() => {
    res.redirect('https://red-badger.com/assets-honestly/551381/19AE4B7538BC1E1C8.css');
  }, 5000)
})

app.listen(3000, function () {
  console.log('Example app listening on port 3000!')
})
```
2. Start the node express app
3. In index.ejs of website-honestly, replace the url in the stylesheet link from `https://cloud.typography.com...` to `http://localhost:3000/` (the node express server)
4. Start website-honestly

**Actual result:**
The content isn't displayed until after 5 seconds

**Expected result:**
The content is displayed straight away

grahammendick:
I think we should follow [Filament Group’s solution](https://github.com/filamentgroup/loadCSS) to asynchronously load css. The content is displayed straight away using the fallback default font specified. When the new font downloads the display starts using it. This changing of default to new font causes a slight flash. Need to check that the UX team are happy with this.

gregdorward:
@sarigriffiths, I think this would be a good one to get near to the top of our backlog, as it should have a good impact on page load performance. It will need UX sign off once it's been put in place, as it's possible the user will see a flash of unstyled text on first load. Ultimately, it's going to be a case of weighing up whether the quicker load time is worth this risk

Issue #434, opened 2017-04-13
Squarespace outage brings down website-honestly


grahammendick:
Squarespace went down yesterday (12/4). The blog part of our site stopped working, as expected. But what wasn’t expected was that website-honestly also went down.

After explaining why Squarespace brought down website-honestly, I’ll suggest what we should do about it to stop it happening again.

## Problem
We use Typography.com for our web fonts. On website-honestly we have a stylesheet link to Typography. Typography decides what font to load based on the browser. It then redirects to the relevant font stylesheet held on our Squarespace server.   

Typography and Squarespace are both involved in serving up our web fonts. If either of them fails then the web fonts won’t be returned. If they fail fast then this isn’t a problem for website-honestly. Website-honestly will work just fine with the default fonts that are built into the browser.

The problem only happens if they fail slowly. This is what happened yesterday. It took Squarespace over 7 seconds to return a failure message. Failing slowly brings down website-honestly because stylesheets are blocking requests. The browser waits until the stylesheet returns before showing any content. If Squarespace takes over 7 seconds to return the font stylesheet then the user won’t see any content at all for that time.

## Solution
The solutions isn’t to move the fonts from Squarespace to website-honestly. This wouldn’t solve the problem. If Typography takes 7 seconds to fail then we’d have the same problem as yesterday.

The solution is to make the stylesheet request from website-honestly non-blocking. That way it won’t matter how long Typography or Squarespace take to respond. Website-honestly will load up straight away with the default fonts and then, when the font stylesheet returns, it will start using the Typography fonts.

I’ve tested out this approach. I throttled the Typography request for 5 seconds and website-honestly showed a blank page for that time. Then I changed the stylesheet link to be non-blocking and website-honestly showed all the content instantly with the default fonts and, after 5 seconds, the fonts updated.

[Filament Group’s solution](https://github.com/filamentgroup/loadCSS) is the recommended approach for asynchronously css loading

gregdorward:
So, if I'm understanding this correctly, on the odd occasion where typography takes the longest to load when hitting our site, the fonts will update after the user's on the page? It's definitely preferable to having an unnecessary dependency on typography, but I'm wondering if that would look a bit strange if it's noticeable.

Is there a 3rd option where we could possibly style to blog extract on the Homepage outside of typography, removing the call entirely from this page?

grahammendick:
[...]

PR #455, opened 2017-05-09
load fonts from typography.com asynchronously

dstodolny:
### Motivation

- [Related story 1](https://github.com/redbadger/website-honestly/issues/438)
- [Related story 2](https://github.com/redbadger/website-honestly/issues/434)

### Test plan

```
git clone https://github.com/dstodolny/typography-com-test
cd typography-com-test
npm i
node typography.js
```
In index.ejs of website-honestly, replace the url in the stylesheet link from https://cloud.typography.com... to http://localhost:3000/ (the node express server)

```
<link rel="preload" href="http://localhost:3000/" as="style" onload="this.rel='stylesheet'" />
<noscript><link rel="stylesheet" href="http://localhost:3000/"></noscript>
```
 
See if it loads fallback fonts first and then after 5 seconds loads correct fonts.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


lpil:
This'll result in the fonts popping in after the user has started reading the content, which is quite jarring. Worth doing user testing with?

gregdorward:
Yeah, @lpil it's something that I intend to get the UX team across. I'm expecting performance gains, but we want to make sure those benefits outweigh the potential UX consequences.

@dpiatek if we decide that the experience is too jarring, is it possible to pull the fonts from the two different sources (removing the dependencies), without the asynchronous loading?

lpil:
The Guardian caches their fonts in localstorage and only makes the request if it's not there. That could be another option.

gregdorward:
looping @grahammendick into the discussion, as he did a fair bit of investigation into how we use Typography, and can probably answer your question @lpil (providing he has the time to)

grahammendick:
It’s FOIT (flash of invisible text - current behaviour) vs FOUT (flash of unstyled text - proposed new behaviour), right? If Typography takes a long time to respond then the flash of invisible text can be much longer than just a flash and can even cause an outage, #434. (Caching in local storage won’t solve the the slowness/outage on first visit.)

gregdorward:
@romanschejbal when you've got a chance, I think this one still needs a review. We've tested most of it, but we ultimately need to leave it to the UXD team to make a call on whether we push it to production

PR #461, opened 2017-05-17
show 'not supported' message in ie11

dstodolny:


gregdorward:
This is looking good @dstodolny - could we get it applied to the Blog as well, and the old site if possible (this part is less of a priority though)

PR #463, opened 2017-05-18
change copy

dstodolny:


gregdorward:
I've approved the code, but just need to check it on the front-end before merging

PR #468, opened 2017-05-22
fixed bottom padding on meet us page

chris-brookes:


gregdorward:
Looks good across browser and devices, both js and non-js, so happy to merge this one

PR #473, opened 2017-05-26
Track footer links with ga

romanschejbal:
### Motivation

- [Related story](https://trello.com/c/x0bhU1xJ/62-add-event-tracking-code-to-navigation-in-footer)

This PR now moves away from the analytics.js script to a React oriented wrapper: react-ga.
This allows us to trigger events easily within the React DOM.

Used to track the usage of links in the footer. Each click on a footer link now triggers an `event` for GA. The usual navigation GA hooks are still there.

### Test plan

1. Push to production, click the footer links.
2. Document the order in which you click the links in footer. (may be useful if anything goes wrong)
3. Check current GA navigations are still being flagged.
4. Wait 24 hours. GA doesn't publish these events in complete real time (I _think_), so you might have to wait 24 hours for the results to show up.

Other edge cases to note:

- Need to make sure the event is fired correctly when navigating to the blog from the footer. The [Google Tag Assistant](https://chrome.google.com/webstore/detail/tag-assistant-by-google/kejbdjndbnbjgmefkgdddjlbokphdefk?hl=en) plugin for chrome is very handy for testing this.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


romanschejbal:
@grahammendick that's for page view, but we need to track if user clicked a link in the footer.. Also, Is it possible to extend the NavigationLink component's `onClick` handler? It seems to override it. Or any other ideas? Can't leave it on the `li` element due to linter rules.

grahammendick:
Use the `navigating` prop https://github.com/redbadger/website-honestly/blob/f7b21ea17c8f21cdccb982a08896922d0378459e/site/components/header/small-screen-nav.js#L83

site/components/footer/index.js
```diff
@@ -72,18 +85,17 @@ const Footer = () => (
...
+            Find us on social media
+          </span>
           <ul className={styles.socialLinks}>
             <li>
               <a
                 href="https://github.com/redbadger"
                 title="Red Badger Github"
                 className={styles.socialIcon}
               >
-                <InlineSVG
```
Xanndarr: Prettier ^

site/components/footer/index.js
```diff
@@ -20,31 +21,43 @@ import mapSVG from './SVG/map.svg';
...
-              <Link to="aboutUsPage">About us</Link>
+              <Link to="aboutUsPage"><span onClick={trackAnalytics('About us')}>About us</span></Link>
             </li>
             <li>
-              <Link to="whatWeDoPage">What we do</Link>
+              <Link to="whatWeDoPage"><span onClick={trackAnalytics('What we do')}>What we do</span></Link>
             </li>
             <li>
-              <a href="/blog/">Blog</a>
+              <a href="/blog/"><span onClick={trackAnalytics('Blog')}>Blog</span></a>
```
JoeStanton: The browser will cancel AJAX requests when natively navigating between pages. This means you might miss the `/blog/` nav event (depending on timing). The others are probably ok as the navigation is tracked client side. Has this been tested? Can use the chrome GA debugger.

site/components/footer/index.js
```diff
@@ -198,12 +193,14 @@ const Footer = () => (
...
           </p>
           <p className={cx('afterDivider', 'disclaimerParagraph')}>
             Registered in England No. 7242017
           </p>
-          <p className={styles.disclaimerParagraph}>VAT Registration No. 990 8085 82</p>
+          <p className={styles.disclaimerParagraph}>
+            VAT Registration No. 990 8085 82
+          </p>
           <p className={styles.cookieWarning}>
             We use cookies on our website. For more information, view our privacy policy.
```
JoeStanton: It would be best to run prettier in bulk so the diffs don't show up in PRs. Can we do this after merge?

site/client/index.js
```diff
@@ -28,6 +29,10 @@ export function makeApp({ element, state }) {
     });
   }
 
+  // init Google Analytics tracker and publish a page view at '/'
+  ReactGA.initialize('UA-16654919-1');
```
JoeStanton: I do prefer this pattern, but it means we'll be initialising this script quite a lot later than we were before. This is probably ok but should keep an eye on it.

site/components/footer/index.js
```diff
@@ -20,31 +21,43 @@ import mapSVG from './SVG/map.svg';
 
 const cx = classnames.bind(styles);
 
+const trackAnalytics = title => () => ReactGA.event({
+  category: 'FooterNavigation',
+  action: title,
+  label: 'click',
+});
```
JoeStanton: Would this work for open in new tab etc?

site/client/index.js
JoeStanton: Its also nice if we can pass this in, maybe an ENV var?

site/components/footer/index.js
Xanndarr: Will set up another PR to do this

site/components/footer/index.js
Xanndarr: Seems to work, yep

site/client/index.js
Xanndarr: Will move over to env var for this

site/components/footer/index.js
Xanndarr: Resolved in #480

site/components/footer/index.js
Xanndarr: Resolved in #480

site/components/footer/index.js
Xanndarr: Tested with Google Tag analyser for chrome. Might want to double check because blog isn't served locally. So it may only be doing half a redirect then crashing. @gregdorward something to watch out for when testing. I will add to the test plan

site/components/footer/index.js
Xanndarr: This works on Cmd+Click and should be the same for Ctrl+click (for windows), but doesn't work with middle click or right click context menu. Will look into a possible fix. Might be better just to put this out and set up a bug for this so we don't forget about it?

grahammendick: Use the `navigating` prop https://github.com/redbadger/website-honestly/blob/f7b21ea17c8f21cdccb982a08896922d0378459e/site/components/header/small-screen-nav.js#L83 instead of span onclick
This handles all those things like right clicking context menu

JoeStanton: OK - I think its probably not worth doing. But the limitation is worth highlighting to whoever will be looking at analytics.

site/components/footer/index.js
JoeStanton: The normal, awful, hacky way to ensure you track the event is to `preventDefault`, then `setTimeout` and then navigate with `window.location` after 200ms or so. Its probably ok to not fix this.

site/components/footer/index.js
Xanndarr: Oh i'll give that a quick go @grahammendick cheers

Xanndarr: Navigating actually broke the tracking :S I'll leave it as is for now and i'll bug it. For now, this is a big improvement. It's also blocking another PR, so getting this through ASAP would be great

PR #483, opened 2017-06-07
Adds window pathname to label instead of click

Xanndarr:
Makes GA label more useful for marketing. We can now see which page users were on when they clicked on a footer link

### DIFF
--- a/site/components/footer/index.js
+++ b/site/components/footer/index.js
@@ -21,20 +21,17 @@ import mapSVG from './SVG/map.svg';
 
 const cx = classnames.bind(styles);
 
-const trackAnalytics = title => () => ReactGA.event({
-  category: 'FooterNavigation',
-  action: title,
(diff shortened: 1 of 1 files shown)


site/components/footer/index.js
```diff
@@ -21,20 +21,17 @@ import mapSVG from './SVG/map.svg';
...
-const trackAnalytics = title => () => ReactGA.event({
-  category: 'FooterNavigation',
-  action: title,
-  label: 'click',
-});
+const trackAnalytics = title => () =>
+  ReactGA.event({
+    category: 'FooterNavigation',
+    action: title,
+    label: `From: ${window.location.pathname}`,
```
Xanndarr: The only change:
```label: 'click' => label: `From: ${window.location.pathname}` ```

The rest is prettier nonsense

PR #489, opened 2017-06-12
5. Tech Page - Open external links in new tabs

Xanndarr:
### Motivation

Open external links in new tabs. Adds `rel="noopener noreferrer"` to protect against SOP vulnerabilities


gregdorward:
- [ ] Update copy in Tech Roundtable slice to be "Technology does not stand still and neither do we. On a regular basis we get our engineering team together to discuss what’s been used on our projects and what emerging technologies could be taken to enterprise scale."

PR #522, opened 2017-08-03
Added 'status' parameter to formatFormInput function

matt-paul:
...to allow different status in post and update member functions.  This will stop multiple confirmation emails being sent.

### Motivation

- [Related story](http://...)

TODO: write a motivation.

### Test plan

- TODO: Write a test plan.

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


chris-brookes:
More changes still to be done.

chris-brookes:
@callerc1 - Martin has done a number of changes to this PR since the last review. Can you give it another quick review please.

services/mailchimp/config.js
```diff
@@ -1,5 +1,5 @@
 const groups = {
-  techPageSignup: '17dfd6ce16',
+  techPageSignup: process.env.ENVIRONMENT_NAME === 'live' ? '17dfd6ce16' : 'd2e5a52843',
```
callerc1: Not sure if this is the correct env var we should be using. Like is this actually an env var in the lamda process? May be worth double checking how this works. I have some ideas around some alternative approaches for the mailchimp group mapping. Let have a chat when you get around to it.

services/mailchimp/config.js
```diff
@@ -1,5 +1,5 @@
 const groups = {
-  techPageSignup: '17dfd6ce16',
+  techPageSignup: process.env.MAILCHIMP_TECH_GROUP_ID,
```
callerc1: way better solution imo 👍

PR #536, opened 2017-08-16
 "Upgrade to webpack 2" v2

chris-brookes:
Reverts redbadger/website-honestly#535

This is to get a copy of the original webpack 2 updates.


chris-brookes:
This PR is no longer needed.

PR #633, opened 2017-12-13
Webinar registration service

frigus02:
### Motivation

We want to replace the current iframe modal for webinar registrations with a custom form, which will be much easier to use.

This is the integration with the GoToWebinar API. It adds a new service, which will accept HTTP POST requests and register the specified user in the specified webinar. It will later be called from our registration form.

This introduces the following new environment variables:

- `GOTOWEBINAR_CLIENT_ID`: The GoToWebinar OAuth 2 client id, which we generated in our GoToWebinar account.
- `GOTOWEBINAR_USER_ID`: The GoToWebinar user name we use to register new participants. It has to have access to the managed webinars.
- `GOTOWEBINAR_PASSWORD`: The password of the GoToWebinar user.

### Test plan

Manually call the serivce. A call will look like this:

```http
POST /dev/webinar-registration HTTP/1.1
Host: https://7l3k9m4j1i.execute-api.eu-west-1.amazonaws.com
Content-Type: application/json

{
    "firstName": "John",
    "lastName": "Doe",
    "email": "john.doe@example.com",
    "webinarKey": "1234567891234567890"
}
```

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review


davidbasalla:
This needs a rebase I think?

davidbasalla:
As an aside, I'm not used to seeing those merge commits (https://github.com/redbadger/website-honestly/pull/633/commits/8144128e5064c7e7eafdabaf22c69c8c7e31db74 and https://github.com/redbadger/website-honestly/pull/633/commits/6e509eac0122232952da79efbbd11bfc72b3d69a), personally I prefer to rebase against master to have a cleaner git history, but not sure what the house rules/preferences are...

markholland:
@davidbasalla for a feature branch it doesn't really matter imo, also with the level of automation going off of commit hashes I'd personally avoid any history rewriting.

services/webinar/authenticate.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+      const credsManager = new ApiCredentialsManager();
+
+      return expect(credsManager.getApiCredentials()).to.eventually.deep.equal({
+        accessToken: 'token',
+        organizerKey: 'organizer',
+      });
+    });
+
+    it('uses cashed creds', () => {
```
davidbasalla: small typo, 'cashed' -> 'cached'?

services/webinar/authenticate.test.js
```diff
@@ -0,0 +1,79 @@
...
+        accessToken: 'token',
+        organizerKey: 'organizer',
+      });
+    });
+
+    it('requests new creds when cached ones are expired', () => {
+      expectHttpRequest(200);
+
+      const credsManager = new ApiCredentialsManager();
+      credsManager.storedAccessToken = 'token';
```
davidbasalla: Minor comment - the `storedAccessToken` and `storedOrganizerKey` could be renamed to something like `old_token` to make it more obvious that the old values don't get used for this test. Otherwise this test does not look very different from the previous one (which tests for using cached values).

services/webinar/api.js
```diff
@@ -0,0 +1,45 @@
...
+  email: string,
+  organizerKey: string,
+  webinarKey: string,
+  accessToken: string,
+}) {
+  const url = `${apiBase}/G2W/rest/organizers/${organizerKey}/webinars/${webinarKey}/registrants`;
+  const query = '?resendConfirmation=true';
+
+  const headers = new Headers();
+  headers.append('Accept', 'application/vnd.citrix.g2wapi-v1.1+json');
```
davidbasalla: What is the 'application/vnd.citrix.g2wapi-v1.1+json' bit for?

services/webinar/api.js
```diff
@@ -0,0 +1,45 @@
...
+  return fetchWithBody(url + query, {
+    method: 'POST',
+    headers,
+    body: JSON.stringify({
+      firstName,
+      lastName,
+      email,
+    }),
+  }).then(({ response, responseBody }) => {
+    if (response.status === 201 || response.status === 409) {
```
davidbasalla: We could add some constant definitions for these status codes (eg `created` for 201 and `conflict` for 409) so that it's easier to parse at first glance. I admit I had to look up what those status codes meant 🙄  I assume that these are the only status codes that we care about?

services/webinar/api.test.js
```diff
@@ -0,0 +1,85 @@
...
+import nock from 'nock';
+
+import { apiBase } from './utils';
+import * as api from './api';
+
+function expectHttpRequest(status) {
+  const reqheaders = {
+    accept: 'application/vnd.citrix.g2wapi-v1.1+json',
+    authorization: 'token',
+    'content-type': 'application/json',
```
davidbasalla: Does 'content-type' need be in single quotes here?

markholland: Yeah I was looking at this, couldn't find a dissallow rule for kebab case

services/webinar/authenticate.test.js
markholland: fixed

services/webinar/api.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
davidbasalla: We could rename this file/module from 'api' to 'register' or similar as that seems to be what it's doing and would be more intention revealing than the current name?

services/webinar/utils.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import fetch, { RequestInit } from 'node-fetch';
+
+export const apiBase = 'https://api.getgo.com';
+
+export function fetchWithBody(
```
davidbasalla: I'm curious about this having to be a custom function? It seems like a pretty standard thing to want in terms of functionality... or is it the use of Promises that makes this 'special'? 
Disclaimer: I'm somewhat new to the wider JS ecosystem so not sure at all about what to expect from the language :)

davidbasalla:
Looks good to me, codewise! I've added a few comments, probably mostly stemming from my ignorance around all things JS... 😄

services/webinar/api.js
frigus02: Needed to tell GoToWebinar to use version 1.1 of the API. Described here: https://goto-developer.logmeininc.com/content/gotowebinar-api-reference

services/webinar/authenticate.js
```diff
@@ -0,0 +1,67 @@
...
+
+import { Headers } from 'node-fetch';
+import querystring from 'querystring';
+
+import { fetchWithBody, apiBase } from './utils';
+
+function parseExpirationDate(expiresIn: string): number {
+  const secondsFromNow = parseInt(expiresIn, 10);
+  const millisFromNow = secondsFromNow * 1000;
+  const expirationDate = Date.now() + millisFromNow;
```
zoeabryant: We use moment in other parts of the app, it might help here. If not, we should have a test to cover this code.

services/webinar/authenticate.js
```diff
@@ -0,0 +1,67 @@
...
+          this.storedOrganizerKey = responseBody.organizer_key;
+          this.storedCredsExpirationDate = parseExpirationDate(responseBody.expires_in);
+
+          return {
+            accessToken: this.storedAccessToken,
+            organizerKey: this.storedOrganizerKey,
+          };
+        }
+
+        throw new Error(JSON.stringify(responseBody));
```
zoeabryant: I can't remember - do we add context to errors thrown to help debugging? Just so we know where this error is being thrown?

services/webinar/authenticate.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  nock(apiBase, { reqheaders })
+    .post('/oauth/access_token', reqbody)
+    .reply(status, resbody);
+}
+
+describe('webinar-registration-service/authenticate', () => {
+  process.env.GOTOWEBINAR_USER_ID = 'USER';
+  process.env.GOTOWEBINAR_PASSWORD = 'PASSWORD';
+  process.env.GOTOWEBINAR_CLIENT_ID = 'CLIENT';
```
zoeabryant: potentially dumb question - do these stay as these values after you've run the test?

services/webinar/utils.js
```diff
@@ -0,0 +1,17 @@
+// @flow
```
zoeabryant: could we change the name of this file to be more specific?

services/webinar/utils.js
```diff
@@ -0,0 +1,17 @@
...
+
+import fetch, { RequestInit } from 'node-fetch';
+
+export const apiBase = 'https://api.getgo.com';
```
zoeabryant: What's the benefit of separating out `apiBase`?  `url` in register_participant.js is very specific

services/webinar/utils.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+      return expect(
+        utils.fetchWithBody('https://example.com/users', {
+          method: 'POST',
+          body: JSON.stringify({
+            username: 'pgte',
+            email: 'pedro.teixeira@gmail.com',
+          }),
+        }),
+      ).to.eventually.nested.include({
```
zoeabryant: ooh, not seen this before, nifty!

services/webinar/authenticate.test.js
zoeabryant: discussed - pass them in as default variables

services/webinar/register_participant.js
```diff
@@ -0,0 +1,50 @@
...
+      firstName,
+      lastName,
+      email,
+    }),
+  }).then(({ response, responseBody }) => {
+    if (response.status === 201 || response.status === 409) {
+      return responseBody;
+    }
+
+    throw new Error(JSON.stringify(responseBody));
```
zoeabryant: same as above ^ add some context for future devs debugging

services/webinar/authenticate.js
zoeabryant: discussed - not worried about using moment as we also use date-fns. Do need the test though

services/webinar/authenticate.js
zoeabryant: discussed - please add context, and to other errors thrown in the service

services/webinar/utils.js
zoeabryant: benefit - used in 2 places!

PR #644, opened 2017-12-15
Add slice to our work page for FM Digital Transformation

davidbasalla:
### Motivation

We add a new slice for the [Fortnum and Mason Digital Transformation Case Study](https://red-badger.com/our-work/case-study/fortnum-and-mason-digital-transformation/) so that users can navigate to it via the 'our work' page.

I mostly copied the code from the Sky CMS slice, as that matched the requested style.

### Test plan

- Visit https://www-staging.red-badger.com/a299987/our-work/, check that the text is correct and that it links to the correct case study (F&M Digital Transformation)

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


site/pages/our-work/case-study-fortnum-and-mason-digital-transformation-slice/index.js
```diff
@@ -0,0 +1,28 @@
...
+            <img src={fmLogo} className={styles.clientLogo} alt="Fortnum and Mason logo" />
+            <h2 className={styles.caseStudyTextContainerHeader}>
+              The ongoing digital transformation of a 310-year old retailer
+            </h2>
+            <p className={styles.description}>
+              Discover how we helped customers find products faster and drive sales of tea.
+            </p>
+            <div className={styles.links}>
+              <p className={styles.readmore}>Read more</p>
+            </div>
```
zoeabryant: if this is an interactive item, could we make it a `button` or an `a` tag? Semantic HTML is important for screenreaders and users who navigate using the keyboard

frigus02: The whole thing is wrapped inside a `<Link />` tag, which is rendered as an `<a />` tag, so I think this sould be fine.

frigus02: Or would it be better for accessibility to wrap each paragraph in a separate link instead of the whole thing? I don't know, to be honest.

site/pages/our-work/case-study-fortnum-and-mason-digital-transformation-slice/style.css
```diff
@@ -0,0 +1,95 @@
+@value badgerBlack, badgerWhite, badgerRedOnWhite, linesOnWhite from "../../../css/_colors.css";
+@value mediumScreen, largeScreen from "../../../css/_sizes.css";
+
+.caseStudyContainer {
```
markholland: Shouldn't these generic case study styles be in one place and then imported?

site/pages/our-work/case-study-fortnum-and-mason-digital-transformation-slice/index.js
zoeabryant: Screenreaders take anything they identify as a link out of context and into one link. So if it just read out this block, nothing else, would it make sense?
I don't think we should do separate links, because 'Read more' by itself doesn't make any sense.

zoeabryant: I just had a listen, I think it'll be ok as is!

site/pages/our-work/case-study-fortnum-and-mason-digital-transformation-slice/style.css
davidbasalla: Yeah I definitely agree. Having looked at the other slices, they all have their own style sheets and annoyingly they have diverged a little bit so it would not be straight forward to extract generic rules for this. We should do this clean up, but maybe as a separate PR as it will touch many other files.

PR #681, opened 2018-01-25
Events banner set from Prismic

markholland:
### Motivation

On the events page we have a static image with a link. We currently have to manually update this and redeploy.

This PR supports a new custom type in Prismic called events-banner that allows setting the link, altText and images for the banner.

### Test plan

- Take a look at the new custom type _Event Banner_ in Prismic staging along with the single instance of it.
- Go to the [events page](https://www-staging.red-badger.com/e8b4adf/events) and observe that the events banner is present with correct link, alttext and images
- Try changing the content of the Events Banner using staging Prismic

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


markholland:
builds will fail until new badger brain is in staging

site/pages/events/index.js
```diff
@@ -1,15 +1,22 @@
...
+import React, { PropTypes } from 'react';
 import Container from '../../components/container';
 import Section from '../../components/section';
 import styles from './style.css';
 import EventsList from './events-list';
 import EventsBanner from '../../components/events-banner';
 
-export default function Events({ events }) {
+export default function Events({ events, eventsBanner }) {
+  const { url, altText, desktopURL, tabletURL, mobileURL } = eventsBanner[0];
```
davidbasalla: Possibly pedantic but it feels a bit odd to have `eventsBanner` as an array here (since the naming implies a singular instance). Instead we could deal with taking the first (and only) entry from the array in 'state/index.js'... up to you though, I don't feel strongly about it since it's only being used here.

markholland: done

chris-brookes:
@dpiatek Little change:
1. Change name of field in Prismic, from 'title' to 'Name', to match the name on the list page.
screen shot 2018-02-27 at 14 13 25

2. Do you think we should keep the old hard coded one to have a backup image if the Prismic one does not exist/appear or is invalid etc?

PR #696, opened 2018-03-06
New innovation and delivery header

joaquimadraz:
### Motivation

[Trello card ](https://trello.com/c/YmMO8tXX/95-what-we-do-innovation-delivery)
Add new innovation and delivery header.

### Test plan

- Navigate to the "what we do"
- New header should show up

### Pre-merge checklist

- [ ] Documentation
- [ ] Unit tests
- [ ] Reviews
  - [ ] Code review
  - [ ] Design review
- [ ] Manual testing
  - [ ] Windows 7 IE 11
  - [ ] Windows 10 Edge
  - [ ] iOS 9 Safari
  - [ ] Android 6 Chrome
- [ ] Tester approved


site/pages/what-we-do/header-slice/style.css
```diff
@@ -0,0 +1,141 @@
...
+  margin-bottom: 20px;
+}
+
+.subtitle {
+  composes: boldSansSerif from "../../../css/typography/_fonts.css";
+  font-weight: bold;
+  text-align: center;
+  line-height: 1.25;
+  margin-top: 20px;
+  font-size: 16px;
```
matt-paul: wondering if you could compose from a preexisting class within fonts?

site/pages/what-we-do/header-slice/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import styles from './style.css';
+import arrowImg from './arrow.png';
+
+const headerSlice = () => {
+  return (
+    <div className={styles.pageHeaderContainer}>
```
matt-paul: perhaps a <section> would be marginally better semantically / for accessibility?

joaquimadraz: Make sense 👍 It also keeps the consistency with the other slices.

site/pages/what-we-do/header-slice/index.js
```diff
@@ -0,0 +1,35 @@
...
+import arrowImg from './arrow.png';
+
+const headerSlice = () => {
+  return (
+    <section className={styles.pageHeaderContainer}>
+      <div className={styles.left}>
+        <div className={styles.flexible}>
+          <h1 className={styles.pageHeader}>
+            <p>Do the</p>
+            <p>right thing</p>
```
dpiatek: This is invalid HTML. `p` tags can only be used where flow content is expected. For an `h1` only phrasing content can be used https://html.spec.whatwg.org/dev/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements

site/pages/what-we-do/header-slice/index.js
```diff
@@ -0,0 +1,35 @@
...
+          </h1>
+          <img role="presentation" className={styles.arrowDown} src={arrowImg} />
```
dpiatek: Using `role` here is good but having `alt=""` is actually better supported https://www.w3.org/WAI/tutorials/images/decorative/

site/pages/what-we-do/header-slice/index.js
```diff
@@ -0,0 +1,35 @@
...
+        </div>
+        <h3 className={styles.pageSubHeader}>Innovation</h3>
+      </div>
+      <div className={styles.right}>
+        <div className={styles.flexible}>
+          <img role="presentation" className={styles.arrowUp} src={arrowImg} />
```
dpiatek: Same comment as above

site/pages/what-we-do/header-slice/index.js
```diff
@@ -0,0 +1,35 @@
...
+          <h1 className={styles.pageHeader}>
+            <p>Do the</p>
+            <p>thing right</p>
+          </h1>
```
dpiatek: As above

site/pages/what-we-do/header-slice/style.css
```diff
@@ -0,0 +1,120 @@
...
+
+.pageHeaderContainer .left {
+  text-align: left;
+  margin-right: auto;
+}
+
+.pageHeaderContainer .right {
+  text-align: right;
+  margin-left: auto;
+}
```
dpiatek: Don't think the descendant selector is needed here (ie. it should be specific enough)?

site/pages/what-we-do/header-slice/index.js
joaquimadraz: Thanks for the resource!

joaquimadraz:
@dpiatek did a couple more corrections.

dpiatek:
:+1:

chris-brookes:
@joaquimadraz bugs:
1. Desktop - bottom sentence font size incorrect. It is 30px, it should be 32px.
2. Alt text missing from the arrows.

screen shot 2018-03-08 at 15 32 47

site/pages/what-we-do/header-slice/style.css
```diff
@@ -112,9 +116,15 @@
 
   .arrowUp {
     margin-right: 60px;
+    transform: rotate(-165deg)
+  }
+
+  .arrowDown {
+    transform: rotate(2deg);
```
dpiatek: Screenshots would be great to explain this change - should we not just replace the asset if we need to do this?

site/pages/what-we-do/header-slice/index.js
```diff
@@ -10,13 +10,21 @@ const headerSlice = () => {
...
+            src={arrowImg}
+          />
         </div>
         <h3 className={styles.pageSubHeader}>Innovation</h3>
       </div>
       <div className={styles.right}>
         <div className={styles.flexible}>
-          <img alt="" className={styles.arrowUp} src={arrowImg} />
+          <img
+            alt="Arrow back up from delivery phrase to innovation phrase."
```
dpiatek: Can we check this alt text with @stevecottle ?

chris-brookes:
[What we do: Innovation + Delivery](https://trello.com/c/YmMO8tXX/95-what-we-do-innovation-delivery)

site/pages/what-we-do/header-slice/index.js
chris-brookes: @dpiatek That's already on our list :-). @joaquimadraz and I spoke about it yesterday.

site/pages/what-we-do/header-slice/style.css
joaquimadraz: The arrow pointing down has a different inclination on the small screen. 
screen shot 2018-03-09 at 10 37 12
screen shot 2018-03-09 at 10 37 01

And it's the same asset for the arrow pointing up so I went for the transformation.

dpiatek: That makes sense 👍

joaquimadraz: Anyway, now that I looked at it, some unnecessary rotations are being done. I'll clean it up.

site/pages/what-we-do/header-slice/index.js
dpiatek: Ok cool

stevecottle: @joaquimadraz @dpiatek What are your thoughts on tagging this up as a whole diagram? For example we put a box around the diagram and have a tag which reads something like 'Cyclic diagram of doing the right thing through innovation and delivery' and ignores the individual components. I'm thinking tagging up the individual components might not be that helpful for those who use screen readers. Also to get a screen reader to focus on each component in a circular fashion might not be possible.

joaquimadraz: @stevecottle 
That makes sense to me. It also reduce the number keystrokes to traverse the page which is good.
I still have to add the attributes for keyboard navigation.

stevecottle: @joaquimadraz Great, let's give it a go. Do you need anything from me?

dpiatek: I think that might be possible with ARIA label attributes, but maybe a simpler solution would to just have the elements as an ordered list? and then the arrow images can again have `alt=""` as order of things can be followed from HTML?

PR #707, opened 2018-03-14
Add best company logo

matt-paul:
### Motivation

- [Related story](https://trello.com/c/sA3j3Mhl/107-add-best-companies-logo-to-job-page)

We are proud of making the Sunday Times 'best small company' list, so would like to display this prominantly on our jobs page.

320px
690px
1600px


site/pages/join-us/style.css
```diff
@@ -11,3 +12,58 @@
 .noWrap {
   white-space: nowrap;
 }
+
+.headerContainer > div {
```
dpiatek: Think we should avoid descendant selectors in our CSS unless it's immediately clear what element it applies to or there is really no other selector we can use.

site/pages/join-us/style.css
```diff
@@ -11,3 +12,58 @@
...
+    display: block;
+    padding-left: 70px;
+    padding-top: 1em;
+  }
+  
+  .headerContainer > picture {
+    display: none;
+  }
+
+  .headerContainer > div {
```
dpiatek: same comment as above

site/pages/join-us/style.css
```diff
@@ -11,3 +12,58 @@
...
+}
+
+@media mediumScreen {
+  .bodyPictureContainer {
+    display: block;
+    padding-left: 70px;
+    padding-top: 1em;
+  }
+  
+  .headerContainer > picture {
```
dpiatek: same comment as above

site/pages/join-us/style.css
```diff
@@ -11,3 +12,58 @@
...
+  margin: 0 auto;
+  width: 120px;
+}
+
+.subHeading {
+  composes: fontS2 from '../../css/typography/_fonts.css';
+  composes: serif from '../../css/typography/_fonts.css';
+}
+
+.headerContainer > div > h1{
```
dpiatek: same comment as above

site/pages/join-us/index.js
```diff
@@ -103,46 +80,84 @@ const apply = {
...
-    <div className={styles.background}>
+class HeaderContainer extends React.Component {
+  static showPicture() {
+    const MEDIUM_BREAKPOINT = '(min-width: 690px)';
+    return !window.matchMedia(MEDIUM_BREAKPOINT).matches;
+  }
+
+  state = {
+    show: HeaderContainer.showPicture(),
+  };
```
dpiatek: For initial state to be set correctly, you need to use a constructor:

```
constructor() {
  super();
  this.state = { show: ... }
}
```

dpiatek: https://reactjs.org/docs/react-component.html#constructor

site/pages/join-us/index.js
```diff
@@ -103,46 +80,84 @@ const apply = {
...
+
+  componentDidMount() {
+    window.addEventListener('resize', this.onResize);
+  }
```
dpiatek: When the component unmounts, we need to remove the listener - or else when the component unmounts, the listener will be kept in memory.
```
componentWillUnmount: function() {
        window.removeEventListener("resize", this.onResize);
    }
});
```

site/pages/join-us/index.js
```diff
@@ -103,46 +80,84 @@ const apply = {
...
   props: {
-    children: 'Current Vacancies',
+    children: <span className={styles.vacancyHeader}>Current Vacancies</span>,
   },
 };
 
-export default function JoinUs({ jobs }) {
-  return (
-    <div className={styles.background}>
+class HeaderContainer extends React.Component {
```
dpiatek: Ideally this component would be in a separate file

matt-paul: moved!

site/pages/join-us/index.js
matt-paul: done, thanks

site/pages/join-us/index.js
matt-paul: happy to do this but i thought this way was ok?  
https://github.com/Khan/style-guides/blob/master/style/react.md#use-es2015-classes
https://stackoverflow.com/questions/35662932/react-constructor-es6-vs-es7
I am probably missing something though, can you elaborate please?

dpiatek: did not know that will work! all good 👍

dpiatek:
@matt-paul can you please rebase this as well (instead of creating a merge commit, as github suggests)

PR #710, opened 2018-03-15
Remove bugsnag fork dependency

dpiatek:
#### Description

Update bugsnag lib - this will remove the only forked repo and simplify yarn install dependecies (ie. no need for git to install packages).

#### Test plan

Errors should still be coming through into bugsnag.


dpiatek:
For reference, original fork was created because of this issue https://github.com/bugsnag/bugsnag-node/issues/90

Issue #691, opened 2018-02-27
Flow usage is inconsistent in the application


dpiatek:
**Description of the issue:**

Flow seems to be added in some places only. Some files use PropTypes or not have the `// @flow` declaration. We should:

- investigate why this is
- replace PropTypes with flow declarations
- add flow where it's missing to increase coverage

dpiatek:
Fixed in #721

PR #721, opened 2018-03-28
Full flow integration

ksunandp:
### Changes
Updating flow to latest version and extending it's integration.
Updating existing flow implementation to reflect latest version.
Fixing any errors caused by update.
Replacing react proptypes with flow.

https://trello.com/c/NMSTGTw6/122-issue-691-flow-usage-is-inconsistent-across-in-the-application
https://github.com/redbadger/website-honestly/issues/691

### Test
Review needed. Passes Circle Tests.

#### Can this PR be merged immediately after testing (eg. rebase, infrastructure update)?
Yes.


site/components/event-image/index.js
```diff
@@ -12,8 +18,9 @@ export default function EventImage({ imgPath, imgAlt }) {
     </div>
   );
 }
-
+/*
 EventImage.propTypes = {
   imgPath: React.PropTypes.string.isRequired,
   imgAlt: React.PropTypes.string.isRequired,
 };
+*/
```
dpiatek: We can remove these now no?

site/components/event-title/index.js
```diff
@@ -1,21 +1,22 @@
...
+// @flow
+import * as React from 'react';
 import styles from '../../pages/events/events-list/style.css';
 import Link from '../link';
 
-const EventTitle = ({ eventLink, eventTitle }) => (
+type EventTitleProps = {
+  eventLink: {
+    to?: string,
+    children?: React.Node,
```
dpiatek: It looks like these might be wrong? Does `Link` work correctly if those values are missing?

dpiatek: It looks like these might be wrong? Does `Link` work correctly if those values are missing?

site/components/header/small-screen-nav.js
```diff
@@ -14,31 +14,36 @@ const trackAnalytics = title => () =>
...
     document.body.style.position = 'fixed';
     this.smallScreenNav.scrollTop = 0;
   };
 
   documentBodyRelease = () => {
+    // $FlowIgnore
     document.getElementById('mainContent').removeAttribute('aria-hidden');
+    // $FlowIgnore
     document.body.style.position = 'relative';
   };
```
dpiatek: Think we can fix this without ignoring see https://github.com/redbadger/website-honestly/blob/circle-2.0/site/components/header/small-screen-nav.js#L31

site/components/header/small-screen-nav.js
```diff
@@ -81,6 +86,7 @@ export default class SmallScreenNav extends React.Component {
           <div className={styles.smallScreenNavMargin} onClick={this.closeMenu} />
           <div
             ref={c => {
+              // $FlowIgnore
               this.smallScreenNav = c;
```
dpiatek: putting `()` around the assignment should help here

site/components/hr/index.js
```diff
@@ -2,12 +2,12 @@ import React, { Component } from 'react';
 import classNames from 'classnames';
```
dpiatek: missing `// @flow` here

site/pages/home/newsletter-slice/index.js
```diff
@@ -61,7 +69,7 @@ class NewsLetter extends Component {
     }
   };
 
-  updateUser = data => {
+  updateUser = (data: any) => {
```
dpiatek: Can we do better here then `any`? It's pretty much a soft bail out from the type system.

site/components/event-title/index.js
ksunandp: I don't know if Link will work with truthfully as no tests associated. My guess would be no, however was no isRequired associated with react prop types. Happy to change though, thoughts?

site/components/header/small-screen-nav.js
ksunandp: The problem is there are null checks and that's why it's falling over. Your file referenced does have those checks. I didn't want to edit any implementation and only the typings. Happy to go back and change?

dpiatek: yup lets change it!

dpiatek: Think there is no avoiding that really with the way the typings work. I'd rather fix it now to be honest.

ksunandp:
Requested changes from review done, commited and pushed.

PR #726, opened 2018-03-29
Upgrade to Circle 2.0

dpiatek:

https://trello.com/c/Vscnf7ce/150-circle-20-update
https://circleci.com/docs/2.0/migrating-from-1-2/

#### Description

Upgrade to Circle 2.0 config as 1.0 is being retired. This will give us a small build time boost as some task will be run in parallel.

#### Test plan

- branch passes
- correct jobs are executed
- branch gets deployed to preview/staging/production


dpiatek:
Note that the build for this _won't go green_ - the branch deletes the 1.0 circle file and build is never run. Merging this will require changing the repo settings (which I can do after review).

PR #741, opened 2018-04-13
Accessibility Day Preparations

matt-paul:
### Description

Add greyscale option to the red badger website to mark accessibility day.  This is a duplication of the code used on the blog for the same purpose.

### DIFF
--- a/site/index.ejs
+++ b/site/index.ejs
@@ -37,7 +37,14 @@
     </script>
     <title><% if (typeof title !== 'undefined') {%><%= title %><% } %></title>
 
+      
     <script>
+
+    /* This script adds the functionality required for accessability day to give the website user an option to toggle greyscale.  
+      It partners with PR #742 (https://github.com/redbadger/website-honestly/pull/742), which holds the UI changes and should be merged on Accessability day.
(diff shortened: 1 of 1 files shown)


site/index.ejs
```diff
@@ -36,6 +36,74 @@
         }
     </script>
     <title><% if (typeof title !== 'undefined') {%><%= title %><% } %></title>
+
+    <script>
+    var colourOnButtonText = 'Make my day accessible';
```
dpiatek: Looks like the indentation is a bit off here

matt-paul: thanks, updated!

site/index.ejs
```diff
@@ -36,6 +36,74 @@
         }
     </script>
     <title><% if (typeof title !== 'undefined') {%><%= title %><% } %></title>
+
+    <script>
+      var colourOnButtonText = 'Make my day accessible';
```
dpiatek: One more thing @matt-paul - could we add a comment here explaining what this is and how it's supposed to work with the other (button) PR and that there is a similar script in blog? Just conscious of people maybe doing this next year and it not necessarily being clear on what is needs to be done.

matt-paul: 👍

PR #744, opened 2018-04-23
Add 2017 data to /about-us 'timeline' slice

matt-paul:
https://trello.com/c/9GLoLuU6/172-update-history-slice-2017-doing-the-right-thing-and-hello-westminster-visual

#### Description
The 'timeline' slice at /about-us was missing information for 2017.  

screen shot 2018-04-23 at 14 37 07


site/pages/about-us/timeline-slice/index.js
```diff
@@ -87,6 +90,9 @@ class TimelineSlice extends React.Component<*, TimelineSliceState> {
               <div className={styles.imageWrapper}>
                 <img src={Image2016} alt="year 2016" className={styles.image} />
               </div>
+              <div className={styles.imageWrapper}>
+                <img src={Image2017} alt="year 2017" className={styles.image} />
+              </div>
               <Six />
```
dpiatek: Think we are missing `Seven` here and the order of components needs to be different too?

dpiatek:
We should remove the outline from the dots and arrows, its not really necessary as we handle the change of state by changing the dot color: 

screen shot 2018-04-26 at 17 04 00

PR #746, opened 2018-04-24
Add hubspot form

dpiatek:
### Changes

- add lambda to subscribe user to hubspot-hosted mailing list
- add lambda to update user data in hubspot mailing list
- amend the forms to support the above

https://trello.com/b/laLbaFJt/internal-projects-website-honestly

#### Test plan

Run `make update-secrets` if testing locally - new ENV variables have been added for hubspot URLS are id's.

The dev environment has staging properties setup in hubspot and will be shared with staging. The office IP is added into hubspot so contacts submitted will be marked as test ones.

Forms in hubspot can be setup to send you a notification when user has submitted and workflow's will show enrolled emails.

#### Are changes needed to infrastructure before/afterwards (eg. deploys to badger brain)?

Production and Staging ENV need to be added to CircleCI as well as corresponding hubspot properties need to be created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ keyrings: ## Initialize blackbox secrets in the keyrings folder (required to get
 	@read -p "Press any key to continue."
 	@$(PRINT_OK)
 
-update-secrets: keyrings ## Update .env file to latest versionea
+update-secrets: keyrings ## Update .env file to latest version
 	cd keyrings \
 	&& git pull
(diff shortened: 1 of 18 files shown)


dpiatek:
This approach does not support double opt-in - leaving this open for now but might we take a wholly different approach to this.

dpiatek:
We are moving the subscribe form to a hubspot landing page, at least till hubspot supports double-opt in from an API call contact submission.

PR #767, opened 2018-05-31
Add microdata from Schema.org to help SEO on events

pataruco:
#### Description

When web spiders crawl sites they take on account semantic HTML to extract data to help search engines, but sometimes they need more info

To tackle this problem Google, Microsoft, Yahoo and Yandex founded [Schema.org](http://schema.org/) to provide consistent information (microdata) for spiders as HTML attributes

I added microdata to event components  from Schema.org to help SEO on events

#### Can this PR be merged immediately after testing (eg. rebase, infrastructure update)?

Yes

#### Test setup (any additional steps that need to be taken to test the story)

Unfortunately, we cannot test this on local environment, because this need to be exposed to web spiders

At the moment when we google it [`red badger events`](https://www.google.co.uk/search?q=red+badger+events&oq=red+badger+events&) we have the following response

screen shot 2018-05-31 at 12 14 16

This display should improve once this code is released

#### Do we need to add any documentation?

I used this [schema](http://schema.org/Event)

### DIFF
--- a/site/components/date-bubble/index.js
+++ b/site/components/date-bubble/index.js
@@ -6,9 +6,10 @@ import styles from './style.css';
 
 export type DateShape = {
   date: string,
+  iso?: string,
+  month?: string,
   monthSym: string,
   year: string,
-  month?: string,
 };
 
 type DateBubbleProps = {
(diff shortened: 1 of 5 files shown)


dpiatek:
@pataruco this looks good. My only issue would be that we don't pass location, which looks like a required prop for the event schema (tested with https://search.google.com/structured-data/testing-tool/u/0/)

PR #847, opened 2019-03-13
Fix image urls in html

sgrowe:
`file-loader` was accidentally being included twice in the "devStatic" webpack config but not the "web" config, causing the urls generated by each set up to differ.


SiAshbery:
:shipit:

PR #853, opened 2019-03-20
Update to node 8

sgrowe:
The lambda [already uses node 8](https://github.com/redbadger/website-honestly/blob/master/services/serverless.yaml#L13) it seems so this PR mainly just updates the CI environment and `.node-version` file


SiAshbery:
:shipit:

PR #868, opened 2019-04-02
Move blocking script to bottom of page

sgrowe:
I noticed there is a synchronously loaded script near the top of the html file which is not great for performance. The aim of the script and the one immediately beneath seems to be to redirect the user to https://red-badger.com/browser-not-supported/ if they're using a version if IE older than 10 (i.e. IE8/9), but given that IE9 browser share is [0.2%](https://caniuse.com/usage-table) it seems silly to harm the performance of the vast majority of browsers, so I've moved it to the scripts to bottom of the page.


SiAshbery:
:shipit:

PR #871, opened 2019-04-09
Responsive images

sgrowe:


SiAshbery:
This is way above and beyond what I expected, great work and thanks for sticking through it, looks like you had a lot of moving parts to deal with.

PR #875, opened 2019-04-25
Add `rel` attribute to link component

reecelucas:
#### Trello Ticket / Github Issue
https://trello.com/c/1HG2GtBN/24-add-rel-attribute-to-cross-origin-links

#### Description
When an anchor has `target="_blank"` is should also have `rel="noopener noreferrer"` to protect against performance and security issues (link to more detail is provided in the Trello ticket). This PR adds a `rel` attribute to anchors rendered by the `Link` component when the `target` prop is set to `"_blank"`.

**Note**: this only solves half the problem, since a fair number of links are injected into the markup from HubSpot. I'm going to poke around to see if there's a way to customise the markup HubSpot links.


SiAshbery:
Good catch 👍 :shipit:

PR #879, opened 2019-04-30
Make carousel navigator dots clickable

reecelucas:
#### Description
This PR changes the navigator dots from `div`s to `button`s so that they can be used to navigate to the corresponding slide.


site/components/navigator/style.css
```diff
@@ -55,10 +55,13 @@
 }
 
 .item {
+  background-color: transparent;
+  border: none;
   width: 36px;
```
sgrowe: Good shout! Could we just increase the height of the buttons a bit so that they're easier to tap/click? The tap area is quite small vertically by default

Screenshot 2019-04-30 at 15 19 17


```suggestion
   width: 36px;
   height: 36px;
```

reecelucas: Nice catch. I've made them the same height as the arrows now.

PR #895, opened 2019-11-08
Code climate has stopped updating

SiAshbery:


SiAshbery:
I re-added code climate. I have no idea what went wrong.

PR #916, opened 2020-03-03
Cypress proof of concept

gregdorward:
#### Description
Scaffolding and initial UI tests for Website-honestly.

The thinking behind this is that we can build it out over time, prove the stability of it and maybe add it to our CI pipeline if it provides value. This could also double up as a demo suite for some of our clients, who still use tools such as Selenium. For this reason, I'm intending to utilise as much of Cypress' capabilities as I can, while still keeping the suite stable - the stubbing in these initial tests being a good example


gregdorward:
Opening this up to feedback/review. My thought is that if it's deemed mergeable, we should hold off on adding these to the build pipeline until we have trust in the stability, especially given it's slightly over-engineered, e.g. the screenshot comparison functionality, for one

PR #918, opened 2020-03-27
Mbp event page

SiAshbery:
##### TIER 1

_Full support. Test all front-end PR's against these._

_WIN10_

- [x] Chrome
- [x] Firefox
- [x] Edge

_Mac High Sierra_

- [x] Chrome
- [x] Safari

_iOS: iPhone_

- [x] Last 2 major iOS versions

_iOS: iPad_

- [x] Last 2 major iOS versions

_Android_

- [x] Last 3 major Android versions; check with mixture of devices

##### TIER 2

_Functional support; some visual differences allowed. Only test these if it's mentioned in the test plan._

_WIN 7_

- [ ] IE11
- [ ] Chrome

_Mac Sierra_

- [ ] Safari

_iOS: iPhone_

- [ ] +1 major iOS versions compared to Tier 1

_iOS: iPad_

- [ ] +1 major iOS versions compared to Tier 1

_Ubuntu_

- [ ] Firefox
- [ ] Chromium

### Test feedback
- [x] Text "RSVP" on both the RSVP buttons should be centred with same padding applied above and below 
- [x] (Accessibility) Should have for=id of form element in form field labels e.g. `<label for="firstname-text" id="firstname-text-label">First Name *</label>`
- [x] Images of presenters should be smaller and align left in mobile view - at the moment they take up entire width from 689px and lower
- [ ] Images of presenters are square and don't have padding before text on Edge
- [x] Selected state should be different for checkbox and text fields (pending guidance from Reece)
- [x] RSVP button shouldn't enable until all mandatory fields are filled correctly (pending guidance from Reece)
- [x] RSVP button hover state should be a different colour (not the same as the disabled state) (pending guidance from Reece)


SiAshbery:
> Images of presenters are square and don't have padding before text on Edge

I could not reproduce this. Which version and platform was it?

PR #945, opened 2020-12-09
Pago FX case study enhancements

pataruco:
As part of this PR we are doing the following

- [x] Updating the hero banner on PagoFx case study to use a transparent PNG with linear gradient as background
- [x] Amend social media metadata for the page
> Done on previous PR #946
- [x] ~Create an "our work"  slice for `our-work/` page~ 
- [x] Create a "What is next" slice for What is next section


gregdorward:
Looking good @pataruco . Am I right in thinking we still don't have the new image for this AC though?

>  Updating the hero banner on PagoFx case study to use a transparent PNG with linear gradient as background

SiAshbery (approved):

PR #1053, opened 2023-01-17
Direct events to HubSpot

charypar:
Point the main menu link at HubSpot. I may need to add a redirect to the website-proxy as well


charypar:
This is blocked on a [badger-brain](https://github.com/redbadger/badger-brain) issue which I'm not sure how to fix at the moment.

